PR #616, opened 2011-11-01
This patch will prevent some random memory leaking.

thekiur:
The internal closedA object was leaking memory while being totally useless.

It had issues with its size, at random times socket.io stopped removing keys from it,
I found that (Object.keys(closedA).length) was more than 350,000 after running for 10 hours.

Edit: i made some more tests, this really fixes a large memory leak, atleast on node 0.4.8


iamwave007:
So is the current version socket.io able to solve this problem of mem leak after discoonection? Automatically?

PR #688, opened 2011-12-16
Use zlib.gzip instead of child_process.spawn("gzip")

fengmk2:
If zlib module available, use it instead of child_process.spawn("gzip")


airandfingers:
I know this is super old, but can we get this fix applied to the 0.9 branch? 0.9.17 is missing this as well.

Issue #886, opened 2012-05-15
Apply 'destroy buffer size' to websockets transport


martinthomson:
The (formerly undocumented) 'destroy buffer size' configuration option is a very sensible one.  This option, applied to the HTTP-based transports prevents clients from sending very large messages and overloading memory on the server.  If a client sends a message with a body that exceeds this number, the connection is dropped.

This is necessary because the transports buffer the entire request body before processing.  With uncontrolled clients, you have to assume that clients are able to send you any rubbish.  This includes infinitely large messages.

Arguably, the default limit (100Mb) is too large, but that's a different problem.

The websockets transport doesn't have any such protection.  It happily receives, buffers and then attempts to assemble fragments up to 4Gb each (2^32) indefinitely.  I haven't yet tested where this falls over, but I'd be surprised if it didn't.

PR #888, opened 2012-05-16
Adding websockets length constraints for #886

martinthomson:
Issue #886 describes the problem.  In short: socket.IO has a vulnerability whereby websocket clients can send an arbitrary amount of data that is buffered by the server until it runs out of memory.  This pull request adds the length constraints for the websockets transport based on the value of 'destroy buffer size'.

### DIFF
--- a/test/transports.websocket.hybi07-12.parser.test.js
+++ b/test/transports.websocket.hybi07-12.parser.test.js
@@ -258,5 +258,103 @@ module.exports = {
     p.add(getBufferFromHexString(packet));
     assert.ok(gotData);
   },
+  'too large binary message generates error': function() {
+      var length = 100;
+      var p = new Parser(10);
+      var message = new Buffer(length);
+      for (var i = 0; i < length; ++i) message[i] = i % 256;
+      var originalMessage = getHexStringFromBuffer(message);
+      var packet = '82 ' + getHybiLengthAsHexString(length, true) + ' 34 83 a8 68 ' + getHexStringFromBuffer(mask(message, '34 83 a8 68'));
+      
+      var gotError = false;
+      p.on('error', function() {
+          gotError = true;
+      });
+      p.add(getBufferFromHexString(packet));
+      assert.ok(gotError);
+  },
+  'too large text message generates error': function() {
+    var length = 300;
+    var p = new Parser(10);
+    var message = 'A';
+    for (var i = 0; i < length; ++i) message += (i % 5).toString();
+    var packet = '81 FE ' + pack(4, message.length) + ' 34 83 a8 68 ' + getHexStringFromBuffer(mask(message, '34 83 a8 68'));
+
+    var gotError = false;
+    p.on('error', function() {
+      gotError = true;
+    });
+
+    p.add(getBufferFromHexString(packet));
+    assert.ok(gotError);
+  },
+  'too large fragmented text message generates error': function() {
+    var length = 300;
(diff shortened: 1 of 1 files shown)


basarat:
Curious why this was closed. Was this protection not needed, Is the server not susceptible to run out of memory because of this. Thanks :rose:

joefeser:
Thanks Obama

rauchg:
@basarat @joefeser such a protection is probably better to enforce at the edge. If you put socket.io behind another layer (like NginX, which is really common), you would basically have to duplicate that setting when WebSocket support is enabled?

basarat:
Cheers. Thanks for the explanation 🌹

LordMajestros:
@rauchg could you please provide a pointer to the nginx settings for this? I have been searching with no success

wwwtyro:
@LordMajestros Not sure, but you might be looking for `proxy_buffer_size`, see https://gist.github.com/atma/6262642 for an example.

LordMajestros:
Thanks @wwwtyro but `proxy_buffers` and `proxy_buffer_size` are used for storing responses from the proxied server i.e. the nginx backend, they don't affect incoming requests or am I reading that wrong http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_buffer_size?

wwwtyro:
@LordMajestros I think you're probably right (though it might be worth testing to see if it affects both directions). Not sure then; maybe @rauchg can illuminate us. :smile:

viperscape:
@LordMajestros I think you're right, proxy_buffer in nginx is for proxy->nginx. In addition, anything too large for the buffer_size gets written to disk! The other option is to not buffer at all which then streams directly to client but this can have unintended side-effects. 
I'm not sure how to limit client buffer size with a websocket connection in nginx, but if I did I'd assume it would also write to disk, which is what it does for regular client http requests. I think relying on the edge to handle max packet size in a websocket frame is a bad idea for socket.io. I know the websocket spec never defined a max size, which is failure on their part if you ask me (and many others agree) and it was left up to the implementation. 
I vote to reopen this PR/issue or have socket.io provide at least one example of preventing this security issue from any proxy engine, even if it was a high level overview. I don't understand why it was closed to begin with, did this hurt performance in some appreciable way?

LordMajestros:
I wrote a PR to limit the size of the payload here https://github.com/websockets/ws/pull/650 and some bug fixes here https://github.com/websockets/ws/pull/650 (note this is in the underlying https://github.com/websockets/ws library which socket.io uses). The PRs have been merged into websockets so it should be possible to add some options exposing these features in socket.io if the maintainers think it is important enough.

D34THWINGS:
I think this security is mendatory as the client can just send anything he wants by just hacking into the JavaScript console and accessing client code. I got a multiplayer game and and I don't want players to have scripts that crashes the server to end a ranked game earlier when they are loosing. Limiting data from clients is a setting that is available in any HTTP server API so why not in WebSockets ? This surely simple to implement with the help of @LordMajestros

llaakso:
+1

Nerfair:
Socket.io is, based on Engine.io, which in turn uses WS. And new versions of Engine.io already passing maxHttpBufferSize parameter to WS constructor, which is maxPayload actually, and by default it is equal to 10E7 (100,000,000) bytes.

https://github.com/socketio/engine.io/blob/master/lib/server.js#L125

Issue #675, opened 2011-12-02
Allow me to specify port AND hostname on .listen()


andyuk:
Hi,

It's important in my app that it listens on a specific IP address. Having the option to pass the hostname would solve this.

Currently I can only use:
.listen(port)

I am working around this by creating a HTTP server myself.

How about supporting:

var options = {
port: 80,
hostname: 'localhost'
};
.listen(options)

Supporting a config option would mean it could still be backwards compatible.

Thanks,
Andy

PR #966, opened 2012-07-29
Make listen() optionally accept a hostname

DavidChouinard:
This enables Socket.io to accept connections on a specific hostname only.

I realize that it's possible to have this functionality by passing an `http.Server` to `listen()`, but this provides a much obvious and seamless API. I've heard of a few people that run monkey-patched versions of Socket.io just for this functionality. 

This fixes Issue #675.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,4 +1,12 @@
 
+0.9.9 / 2012-08-01
+==================
+
+  * Fixed sync disconnect xhrs handling
+  * Put license text in its own file (#965)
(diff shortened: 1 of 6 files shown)


rmccullagh:
5 years later, and it's extremely challenging for this library to be bound to a specific interface. It's beyond me why such a thing is not documented anywhere on the socket.io website. Perhaps socket.io is just a toy.

PR #1310, opened 2013-09-06
Update namespace.js fix memory leak in acks

nickhsharp:
updating namespace.js similar to socket.io-client/dist/socket.io.js such that it clears the acks object after running the ack function... prevents memory leak.


iamwave007:
http://stackoverflow.com/questions/9918203/remove-objects-on-disconnect-socket-io/32581156#32581156

I'm having doubt over this socket.io memory issue..

PR #1428, opened 2014-02-20
Ability to get all clients for a given room

crickeys:
Not sure if I fully understand all the parts to the new socket 1.0, but I needed this ability. I believe I separated the parts into the namespace and adapter correctly, but please do check me over. Thanks!


akashkrishnan:
:+1:

robbennet:
Why has this not been implemented yet?

PR #1471, opened 2014-03-28
support cors

liorsion:
socket.io supports cors internally today. The problem with this approach is that sometimes socket.io is used with platforms such as express (ie, serving other content) - this content often has it's own way of supporting CORS

For example, a popular setup would be to have the nginx or apache in front of the node.js handle the cors (saving the need to upgrade and offloading the prefetch requests).

In those cases, the socket.io part returns double the CORS headers, and this is illegal in CORS - it fails in Chrome for example.

This change request enables optionally to turn off cors support. By default it's still on

### DIFF
--- a/lib/transports/http.js
+++ b/lib/transports/http.js
@@ -75,7 +75,7 @@ HTTPTransport.prototype.handleRequest = function (req) {
       self.onClose();
     });
 
-    if (origin) {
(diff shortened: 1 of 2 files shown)


buunguyen:
:+1: I'm having the exact problem handling CORS using Nginx. Not sure why the issue is closed.

PR #1596, opened 2014-06-06
Add except support to namespaces to prevent relaying message

gregtap:


buunguyen:
Can't we have both? We can name the method of this PR `excludeClient()` (or `excludeSocket()`) and keep the current implementation. Later, if/when there's the general `excludeRoom()`, `excludeClient()` can be refactored to use `excludeRoom()` without having to change the surface API. 

Would love to see this merged somehow, some people (myself included) are asking for this feature in [this thread](https://github.com/Automattic/socket.io/issues/1595).

bevacqua:
@rase- any update on this? the `exclude` API seems nowhere to be found

PR #1663, opened 2014-07-04
fix cross domain bug when disconnect

lovemyliwu:
fix cross domain bug when disconnect


defunctzombie:
What is the bug? Can you add a failing test?

PR #1693, opened 2014-07-25
Use port 443 for https in Server.checkRequest

evanlucas:
Currently, even if the parsed url.protocol is `https:`,
Server.checkRequest was matching against port 80. This defaults that
check to use 443 if the protocol is `https:`

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -68,7 +68,12 @@ Server.prototype.checkRequest = function(req, fn) {
   if (origin) {
     try {
       var parts = url.parse(origin);
-      parts.port = parts.port || 80;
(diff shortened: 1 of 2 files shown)


defunctzombie:
@rase- is this still an issue?

rauchg:
I went ahead and pushed this myself because it needed a rebase. Thanks @evanlucas

mfn:
I apologize for asking this here, but any reason this didn't end up in any release yet?

Was bitten my this bug, came here to learn it's already fixed but not available in any release (i.e. 1.3.5 does not contain it).

Thank you

PR #1704, opened 2014-08-03
domain access add all subdomain(*.xx.yy) rule

seektan:
also object value check as a substitute for indexOf, avoid this situation: !!"aaa.bb.cc".indexOf("a.bb.cc")

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,4 +1,3 @@
-
 /**
  * Module dependencies.
  */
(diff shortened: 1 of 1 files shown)


defunctzombie:
This PR needs to be cleaned up and the issue described. It also needs tests.

rauchg:
We need a failing test for this

darrachequesne:
Closed due to inactivity. Please reopen if needed.

Issue #1639, opened 2014-06-19
Fixing TypeError: Function.prototype.apply: Arguments list has wrong type


t3knos:
This began to appear in v.1.0.4 and persists in v.1.0.5, and although it __appears__ that my socket.io-client is sending a string that looks like an object, I don't see how I've managed to do that.

```javascript
    /* Client code */
    var getAllRS_sock = io(svr,{
        'multiplex'                 : false,
        'timeout'                   : 5000,
        'transports'              : ['websocket'],
        'reconnection'                 : false,
        'sync disconnect on unload' : true,  // is this still available?        
    });
    /*** When connected, send get_AllRS ***/
    getAllRS_sock.on('connect', function getAllRS_sock_on_connect() {
        console.log('In getAllRS_sock_on_connect()');
        /*** Let's talk to the server ***/
        getAllRS_sock.emit('get_AllRS',{
                                'cType'     : 'get_AllRS',
                                'DEBUG'     : RS_DEBUG,
                                'clientID'  : clientID,
                                'userID'    : userID,
                            }) ;
    });

```

And the server looks like:

```javascript
    /* Server code */
    socket.sockets.on('connection', function socket_on_connection(clientSock){

        webClient = new WebClient(clientSock.id,clientSock.request.connection.remoteAddress,clientSock.request.connection.remotePort);
        logger.warn('+++> Client '+webClient.user+'@'+webClient.addy+':'+webClient.port+" ["+webClient.sess+'] connected ++++++++++++++');    

        clientSock.on('get_AllRS', function clientSock_on_get_AllRS(incomingData){
            logger.warn('*** INCOMING:',incomingData); // We never get this far.
        });
    });
```

Upon the client connecting to the server, the ```DEBUG``` shows the node server crash with the following:

```bash
[2014-06-19 12:13:49.112][8853][0@10.10.17.239:49673 [EaZQ5kULN0yMHRyMAAAA]][ctyp:undef][fnom:undef]
>> [WARN] [emitted] : +++> Client 0@10.10.17.239:49673 [EaZQ5kULN0yMHRyMAAAA] connected ++++++++++++++
[2014-06-19 12:13:49.113][8853][0@10.10.17.239:49673 [EaZQ5kULN0yMHRyMAAAA]][ctyp:undef][fnom:undef]
>> [ERROR] [console] :   socket.io:socket joined room EaZQ5kULN0yMHRyMAAAA +3ms
[...]

PR #1770, opened 2014-09-08
Fix for Issue #1639

steelbrain:
if the stuff recieved is an array as string, it'll be json parsed automatically


defunctzombie:
Incorrect code style.

steelbrain:
:stuck_out_tongue: I don't even remember this pull request

PR #1815, opened 2014-10-07
Chat example using Cluster + Redis Adapter

objectiveSee:
I have forked the chat example and made it work using Cluster.  There is an updated README within the example's directory that instructs you on how to configure Redis.

My solution used the `sticky-session` module (https://github.com/indutny/sticky-session) for routing requests to a slave based on IP address.  This means that if you want to test this application, you will need to hit index.html from different machines in order to see requests hitting distinct slaves.

I unfortunately had to drop support for showing the number of connected users in the chat room. In Socket.IO 1.2.0 there will be support for listing the number of connected clients, and therefore being able to get the count (https://github.com/Automattic/socket.io/pull/1630#issuecomment-55276496). 

I would love some feedback from the project maintainers about how I did on using cluster, and whether my solution is in-line with their vision on the best practice for scaling Socket.io across multiple nodes.

### DIFF
--- a/examples/chat-cluster/README.md
+++ b/examples/chat-cluster/README.md
@@ -1,0 +1,25 @@
+
+# Socket.IO Chat
+
+A simple chat demo for socket.io
+
+## How to use
+
+```
+$ cd socket.io
+$ npm install
+$ cd examples/chat
+$ npm install
+$ node .
+```
+
+And point your browser to `http://localhost:3000`. Optionally, specify
+a port by supplying the `PORT` env variable.
+
+## Features
+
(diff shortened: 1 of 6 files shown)


defunctzombie:
I think this is better as a stand alone example for now. That sticky module does not work perfectly (we have noticed delays in connecting) so am not sure we want to promote that yet. Thanks for working on this but I we will keep our current example.

objectiveSee:
No worries @defunctzombie! I just joined your socket.io cloud channel on slack.  I think we can find some common problems to solve together,

PR #1865, opened 2014-11-15
Add setupNamespace and dynamic namespace support

davidbau:
This pull requests adds dynamic namespace support, virtual host support, and an optional /socket.io/status page to help monitor dynamic (and regular) namespaces.

It addresses https://github.com/Automattic/socket.io/issues/1854

From the library user's point of view, it doesn't change any existing behavior.  Passes current unit tests, but adds some functionality.

#### Three new options.

host: true - enables hostname qualification on namespaces
host: 'roothost.com' - enables hostnames, but matching hosts (may be a regexp) are still unqualified
host: '*' - the default - all hosts are unqualified

retirement: 10000 (default) - the number of ms to wait before cleaning up an empty dynamic namespace

serveStatus: false (default) - set to true to enable socket.io/status

#### One main new public API

server.setupNamespace(pattern, fn) - call to set up all namespaces matching "pattern" (which may be a string or regexp).

The rules are: each namespace instance is called with at most one setup function.  When a new dynamic namespace is connected by a client, the best-matching setup function is found and called.  Exact string matches are preferred over regexps, and within those groups, the last registered name/pattern gets precedence.  If no pattern matches a client's requested namespace name, an error is sent to the client.

#### Other public functions

server.getHost(conn) - override to normalize hostnames
namespace.host - has the hostname
namespace.fullname() - '//host/name'
namespace.retriement - set to milliseconds to persist after going empty (defaulted)
namespace.expire(fn) - register an expiration callback for when a stale namespace is deleted

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -20,9 +20,10 @@ module.exports = Client;
  * @api private
  */
 
-function Client(server, conn){
+function Client(server, conn, host){
   this.server = server;
   this.conn = conn;
+  this.host = host;
   this.encoder = new parser.Encoder();
   this.decoder = new parser.Decoder();
   this.id = conn.id;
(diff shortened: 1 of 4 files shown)


defunctzombie:
Not a fan of this right now. I think it introduces a bit too much magic and custom response pages. I am not against this idea in general, but I somehow feel that regex or dynamic namespaces could be supported in a simpler manner.

Closing since @rauchg mentioned it would be better suited with hooks or similar for now. I don't want to leave it open to get stale.

PR #1893, opened 2014-11-25
Simpler koa integration

iliakan:


defunctzombie:
I think the current example is alright since it shows a unique usage that might not be immediately obvious. Maybe the examples directory could show some variations.

PR #1903, opened 2014-12-08
Added tests for volatile

rase-:
Tests the following behavior with the existing volatility implementation:
- Volatile events can be sent
- Volatile events are not sent after sending a regular event (in the same tick)
- Normal events can be queued normally after a successful (not dropped) volatile event
- Only one volatile should be sent when more than one sent consecutively

All these properties are tested both on the polling and ws transports.


rauchg:
Solid stuff @rase-

Issue #1670, opened 2014-07-08
Unexpected behavior of room leaving


GuoZhang:
I found that in the released 1.0.6,  leaving a room will result in leaving not only the room but also rooms after it. For example, a user is now in room 'A', 'B' and 'C'. Leaving room 'B' will leave the user in room 'A' only (both 'B' and 'C' disappear). 

By checking the code, lib/socket.js, I found:
```js
Socket.prototype.leave = function(room, fn){
  // ... 
  self.rooms.splice(self.rooms.indexOf(room, 1));
  // ...
  return this;
};
```

I guess there is a missing argument (in order to leave the specific room only).
```js
self.rooms.splice(self.rooms.indexOf(room, 1), 1);
```

Hakadel:
:up: 

I have the same issue, do you have a fix for this bug?

PR #1935, opened 2015-01-08
Fixing Bug #1670 - Socket.prototype.leave exhibits wrong behaviour

strikeout:
A bugfix introduced to fix bug #1670 when leaving rooms still exhibits wrong and unexpected behaviour.
If the room that should be left is not found in self.rooms, indexOf returns -1, 
making the arguments of arr.splice (-1,1), meaning remove one room from the end of self.rooms.


defunctzombie:
I added a test. Also, did the fix slightly differently to avoid double call to indexOf. Thanks for pointing this out!

https://github.com/Automattic/socket.io/commit/ca82c09bf21dee5be3f40f0969f2df0cf3d6a7d3

PR #1938, opened 2015-01-11
update license with up-to-date year range

fay-jai:
Only made 1 change:
1) Modified the license to include a range with the current year.

No other changes made.


rauchg:
Awesome! Thanks

fay-jai:
You're welcome!

PR #1940, opened 2015-01-13
Added another example.

heartz:
Its a shared Canvas that I made using Socket.io.
Its still very crude with only basic figures working, but I plan on working on it and add more features on a daily basis.


rauchg:
We think this is better suited to userland code for the time being. We still need to clean up the chat example a lot (especially coding style and best practices).

rauchg:
Really appreciate the contribution!

PR #1941, opened 2015-01-13
Updated main.js

heartz:
Fixed Capitalization in comments


rauchg:
We usually don't capitalize comments I believe. Thanks tho

heartz:
there are multiple comments that are capitalized though, might wanna change them then

PR #1958, opened 2015-01-21
Support compression

nkzawa:
Support compression and add the `compress` method which user can set the `compress` flag for a subsequent event emission.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -341,6 +341,18 @@ server.listen(3000);
   });
   ```
 
+### Socket#compress(v:Boolean):Socket
+
+  Sets a modifier for a subsequent event emission that the event data will
+  only be _compressed_ if the value is `true`. Defaults to `true` when you don't call the method.
+
+  ```js
+  var io = require('socket.io')();
+  io.on('connection', function(socket){
+    socket.compress(false).emit('an event', { some: 'data' });
+  });
+  ```
+
 ### Client
 
   The `Client` class represents an incoming transport (engine.io)
--- a/lib/client.js
+++ b/lib/client.js
@@ -61,7 +61,7 @@ Client.prototype.setup = function(){
 Client.prototype.connect = function(name){
   debug('connecting to namespace %s', name);
   if (!this.server.nsps[name]) {
-    this.packet({ type: parser.ERROR, nsp: name, data : 'Invalid namespace'});
+    this.packet({ type: parser.ERROR, nsp: name, data : 'Invalid namespace'}, false, false, true);
     return;
   }
   var nsp = this.server.of(name);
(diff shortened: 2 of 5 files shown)


nkzawa:
I think `nocompress` flag also can be an option instead of `compress(false)`.

```js
socket.nocompress.emit('myevent');
```

rauchg:
Since I think it's fairly rare, a `.compress(false)` flag should suffice.

package.json
```diff
@@ -19,7 +19,7 @@
     "test": "mocha --reporter dot --slow 200ms --bail"
   },
   "dependencies": {
-    "engine.io": "1.5.1",
+    "engine.io": "Automattic/engine.io",
```
rauchg: Let's point to a specific `#commit`

lib/client.js
```diff
@@ -61,7 +61,7 @@ Client.prototype.setup = function(){
 Client.prototype.connect = function(name){
   debug('connecting to namespace %s', name);
   if (!this.server.nsps[name]) {
-    this.packet({ type: parser.ERROR, nsp: name, data : 'Invalid namespace'});
+    this.packet({ type: parser.ERROR, nsp: name, data : 'Invalid namespace'}, false, false, true);
```
rauchg: We might want to reconsider this signature later on.

Readme.md
```diff
@@ -341,6 +341,18 @@ server.listen(3000);
...
+
+  Sets a modifier for a subsequent event emission that the event data will
+  only be _compressed_ if the value is `true`. Defaults to `true` when you don't call the method.
+
+  ```js
+  var io = require('socket.io')();
+  io.on('connection', function(socket){
+    socket.compress(false).emit('an event', { some: 'data' });
+  });
+  ```
```
rauchg: beautiful

lib/client.js
```diff
@@ -135,17 +135,18 @@ Client.prototype.close = function(){
...
-Client.prototype.packet = function(packet, preEncoded, volatile){
+Client.prototype.packet = function(packet, preEncoded, volatile, compress){
   var self = this;
 
   // this writes to the actual connection
   function writeToEngine(encodedPackets) {
     if (volatile && !self.conn.transport.writable) return;
     for (var i = 0; i < encodedPackets.length; i++) {
-      self.conn.write(encodedPackets[i]);
+      self.conn.write(encodedPackets[i], { compress: compress });
```
rauchg: Just curious, is this the same signature the WS client would use? I mention this since it'd be nice to maintain parity so that you could technically replace `engine.io-client` with a `WebSocket` client altogether.

nkzawa: Yes, `ws` has the same signature, but it's not something formal since WS API for browser has no spec for this.

lib/client.js
nkzawa: yeah, I also don't like this. `options` object would be nice.

nkzawa: I think a packet should have these options, so that we can make it isomorphic with the client.

https://github.com/Automattic/socket.io-client/pull/805/files#diff-9b2f90c89d460ee80ced18e01748824eR139

lib/client.js
rauchg: Right, but that's good enough then…

lib/client.js
rauchg: We can leave this for a different PR @nkzawa

package.json
```diff
@@ -19,7 +19,7 @@
     "test": "mocha --reporter dot --slow 200ms --bail"
   },
   "dependencies": {
-    "engine.io": "1.5.1",
+    "engine.io": "Automattic/engine.io#ddc64a2d1286295fa7beaf2b113d6cbe4fbbc66b",
```
rauchg: Next time you can just use 7 digits.

PR #1961, opened 2015-01-23
Add Server#addListener and Server#removeListener

elierotenberg:
Now we can remove a previously attached main namespace listener directly, eg. removing a `connection` event handler.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -329,7 +329,7 @@ Server.prototype.onconnection = function(conn){
 
 Server.prototype.of = function(name, fn){
   if (String(name)[0] !== '/') name = '/' + name;
-  
+
   if (!this.nsps[name]) {
     debug('initializing namespace %s', name);
     var nsp = new Namespace(this, name);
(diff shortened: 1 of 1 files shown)


pwnsdx:
Bump

pwnsdx:
I set RemoveListener function in the very first line of an event. After a test, I got 2/4 success. It's probably because Node.JS is async. removeListener is therefore not suitable for its usage.

I did not tested addListener as I don't know what it is for (there is already .on)

nkzawa:
I'm not sure whether those shortcut methods are necessary.
I feel we should find a better way if intend to proxy all methods and flags.
Actually there are some others like `once` and `removeAllListeners` as well.

pwnsdx:
Once ? Is there any documentation about that please ?

nkzawa:
http://nodejs.org/api/events.html

Since `Namespace` is actually an `EventEmitter`.

pwnsdx:
Oh, thanks I didn’t know :D

Sabri

darrachequesne:
Closed by https://github.com/socketio/socket.io/pull/2601. Thanks!

PR #1962, opened 2015-01-25
added domain access rule for origins with *.example.com format

alxyang:
includes failing test, checks for subdomains using regex one-liner
supports multiple subdomains e.g. herp.derp.example.com or just herp.example.com
both will return "example.com" as expected


darrachequesne:

PR #1964, opened 2015-01-25
Fixed test semantics.

cha0s:
A couple of tests seemed to be describing functionality different than they were actually testing. I updated the tests to reflect the semantics as described.


cha0s:
I noticed I made a mistake, closing this one and will open another.

PR #1965, opened 2015-01-25
Fixed test semantics, though tests seem to be failing now.

cha0s:
A couple tests seemed to be testing different semantics than what they described, so I updated them. However, upon updating the test `should be equivalent for "" and "/" on client`, the tests appear to be broken. I may be misunderstanding the wording of the unit tests, but it seems that the tests shouldn't fail here if the functionality as described is working.

### DIFF
--- a/test/socket.io.js
+++ b/test/socket.io.js
@@ -467,20 +467,21 @@ describe('socket.io', function(){
       sio.of('').on('connection', function(){
         --total || done();
       });
-      sio.of('abc').on('connection', function(){
+      sio.of('/').on('connection', function(){
         --total || done();
       });
       var c1 = client(srv, '/');
-      var c2 = client(srv, '/abc');
     });
 
     it('should be equivalent for "" and "/" on client', function(done){
       var srv = http();
       var sio = io(srv);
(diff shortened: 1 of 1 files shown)


cha0s:
I found the culprit: we need to use forceNew to force opening a separate socket connection for the 'same' namespace.

rauchg:
We're actually fixing exactly that for 1.4 :) Sorry about that!

cha0s:
It's all good :) It works locally, I guess the bot doesn't like me. I'll close the PR if this is already on your guys' radar.

rauchg:
Actually, tests fixes welcome.

rauchg:
Will keep this open until I merge the fixes for same-namespace different-connection new default.

cha0s:
Rerolled against https://github.com/Automattic/socket.io/issues/1956 & https://github.com/Automattic/socket.io-client#808

cha0s:
Sidenote, that travis bot seems to show that really large binary file test timing out (happens locally too). We should probably bump --timeout in the Makefile to ~5000ms

darrachequesne:

PR #1974, opened 2015-02-01
Suggestion for intercom API

cha0s:
I've found that it would be very useful for synchronization purposes to be able to send messages to server-side sockets in namespaces/rooms directly without transiting to/from the client.

My proposal for this functionality is a new method `intercom` which functions essentially identically to the `broadcast` API, except that it broadcasts to server-side sockets instead of client-side sockets.

`intercom` is implemented as both a `Namespace` instance method as well as a flag for sockets. This allows syntax like:

`io.of('/namespace').in('room').intercom('event', data);`

as well as

`socket.intercom.emit('event', data);`

the latter following the same semantics as its broadcast equivalent -- namely, the origin socket is skipped in the emission.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -292,6 +292,26 @@ server.listen(3000);
   Errors passed to middleware callbacks are sent as special `error`
   packets to clients.
 
+### Namespace#intercom(name:String[, …]):Namespace
+
+  Emits an event to server sockets socket identified by the string `name`. Any
+  other parameters can be included.
+
+  All datastructures are supported, including `Buffer`. JavaScript
+  functions can't be serialized/deserialized.
+
+  ```js
+  var io = require('socket.io')();
+  io.intercom('an event', { some: 'data' });
+  ```
+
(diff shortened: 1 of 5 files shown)


cha0s:
I just noticed `clients` was pulled in so I have to rebase off of that first. Coming shortly.

andreialecu:
Would really like this in socket.io. 

In my scenario, I have a HTTP REST based api that among other things coordinates how sockets join channels. If I'm running on multiple nodes, the session may be connected to node A, while the control API may receive the request on node B. So it can't find the socket session to join it. 

I need to broadcast a message across all nodes to find and join the specific session to a channel. 

To do that, I need to set up a separate redis pub/sub channel among all nodes. 

socket.io-redis already keeps this pub/sub active, so it would be nice to be able to tap into that channel without needing to open two additional redis connections (which with cloud-hosted redis are a limited resource)

darrachequesne:

cha0s:
This isn't an "issue", it's a feature request with a pull request.

PR #1978, opened 2015-02-03
Reset rooms array after each clients() call

Nibbler999:
If you call clients() multiple times with different rooms you end up with clients from all the rooms specified in previous calls.

### DIFF
--- a/test/socket.io.js
+++ b/test/socket.io.js
@@ -730,6 +730,55 @@ describe('socket.io', function(){
         });
       }
     });
+
+    it('should find the correct clients on subsequent runs', function(done){
+      var srv = http();
+      var sio = io(srv);
+      var chatFooSid = null;
+      var chatBarSid = null;
+      var otherSid = null;
+      srv.listen(function(){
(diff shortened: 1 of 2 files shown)


test/socket.io.js
```diff
@@ -730,6 +730,55 @@ describe('socket.io', function(){
...
+        sio.of('/other').on('connection', function(socket){
+          socket.join('foo', function() {
+            otherSid = socket.id;
+            --total || getClients();
+          });
+        });
+      });
+      function getClients() {
+        sio.of('/chat').in('foo').clients(function(error, sids) {
+          expect(error).to.be.undefined;
```
nkzawa: I believe there is no syntax like this on `expect.js`. It should be `expect().to.be(undefined)`.

nkzawa:
:+1: it almost looks good to me.

Nibbler999:
Changed test to use `expect(error).to.not.be.ok()`

rauchg:
Great find.

PR #1999, opened 2015-02-11
Fix a few assertions are not executed properly

nkzawa:


test/socket.io.js
```diff
@@ -636,11 +636,10 @@ describe('socket.io', function(){
       });
       function getClients() {
         sio.of('/chat').clients(function(error, sids) {
-          expect(error).to.be.undefined;
```
rase-: I'm wondering if it's cleaner to just do

```js
expect(error).to.be(undefined)
```

instead.

nkzawa: It doesn't matter either way for me. But the value of `error` is actually `null`, so `.to.not.be.ok()` would be better.

rase-: Ah. Aye.

nkzawa: fixed and rebased.

nkzawa:
umm, why the ci build failed?

nkzawa:
I restarted the build and it's fine now.

PR #2027, opened 2015-03-01
Add helper function Server.has(namespaceId) to check for existance of namespace

jbaumbach:
Allows the client codebase to check for a namespace before setting up an additional listener.  Adding multiple listeners to the same namespace can cause too many 'emit' events to get called.  So, this fork update allows code like this to work:

```
exports.createNamespace = function(namespaceId, storageFunc, cb) {
  //
  // Socket listener.  Gets messages from the clients and rebroadcasts them
  // to all the other clients.
  //

  var ioNamespaceId = '/' + namespaceId;

  if (io.has(ioNamespaceId)) {
    debug('reusing existing namespace: ' + ioNamespaceId);
  } else {
    var nsp = io.of(ioNamespaceId);
    debug('created socket namespace: ' + ioNamespaceId);

    nsp.on('connection', function(socket){
      debug('got connection!');
      socket.on('addComment', function(data) {
        storageFunc(data, function(err, result) {
          if (err) {
            debug('(error) can\'t save data! ' + util.inspect(err));
          }
          socket.broadcast.emit('addedComment', data);
        });
      });
    });
  }
  cb();
};
```

Also - adding additional ".idea" entry to .gitignore so Webstorm users don't have to manually add it when they fork.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ lib-cov
 benchmarks/*.png
 node_modules
 coverage
+.idea
--- a/Readme.md
+++ b/Readme.md
@@ -192,7 +192,7 @@ server.listen(3000);
   ```js
   var io = require('socket.io')();
   var namespaceId = '/mynamespace123';
-  if io.has(namespaceId) {
+  if (io.has(namespaceId)) {
     debug('namespace ' + namespaceId + ' exists!';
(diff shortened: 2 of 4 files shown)


darrachequesne:

PR #2063, opened 2015-03-24
Let adapter.add find connected socket

kapouer:
When a socket connects, it joins its own room, resulting in a call to adapter.add.
The adapter in turn should be able to find the socket by id.


rauchg:
Makes sense.

rauchg:
A comment would be in order considering the big semantic change a line change produced :D

kapouer:
Added in PR https://github.com/Automattic/socket.io/pull/2081

PR #2103, opened 2015-04-30
whitebaord example

rubenabergel:
I created a the simple collaborative whiteboard example with sockets.io.


nkzawa:
Thanks!!! But as far as I tried, this has some problems, like 2 boards don't get the same result when they use different colors.

darrachequesne:

PR #2136, opened 2015-05-29
Don't set request property

nkzawa:
Socket#request is defined as getter, so we shouldn't try to set value.

Related: https://github.com/Automattic/socket.io/issues/2120

### DIFF
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -60,7 +60,6 @@ function Socket(nsp, client){
   this.server = nsp.server;
   this.adapter = this.nsp.adapter;
   this.id = client.id;
-  this.request = client.request;
   this.client = client;
   this.conn = client.conn;
   this.rooms = [];


shybovycha:
I vote up for this pull request. This small patch fixes the whole socket.io when used with iojs and `--use_strict`. 

The request property itself seems meaningless for me, as the request object is contained in `Socket.conn.request` property.

tabone:
:+1:

devoto13:
Node.js 4.0.0 was released 6 days ago, so now this is not just issue for io.js, but for Node.js as well.

Can you give us some update when it's going to be merged? Or is it going to be merged at all?

akotlar:
I vote it up as well.

kevinhikaruevans:
+1

omarloren:
Im using Node v4.1.1 and this issue makes socket.io useless in strict mode, please merge.

devoto13:
For anyone experiencing this issue. Instead of running `node --use_strict` to access ES6 features, which require strict mode, you can run specific files in strict mode by adding `"use strict";` in the first line of those files and run `node` in normal mode. It will allow you to use ES6 features and avoid problems with third-party libraries, which don't support strict mode yet.

darrachequesne:
Hi! Shouldn't this part be updated too?

```javascript
Socket.prototype.buildHandshake = function(){
  return {
    headers: this.request.headers,
    time: (new Date) + '',
    address: this.conn.remoteAddress,
    xdomain: !!this.request.headers.origin,
    secure: !!this.request.connection.encrypted,
    issued: +(new Date),
    url: this.request.url,
    query: url.parse(this.request.url, true).query || {}
  };
};
```
Yet no test seems to fail :innocent:

chatziko:
+1

It's annoying to be unable to --use_strict because of a 1-line bug :(

rauchg:
👌

akotlar:
Thanks!

chatziko:
thanks!

On a related note, is there any reason not to have `"use strict";` in every file? 

node itself uses it, it's compatible, allows optimizations and catches bugs like this one.

rauchg:
@chatziko we're headed there. We'll enable Babel compilation and have that added for us.

Issue #1847, opened 2014-10-31
how to config? io.enable('browser client minification') gone?


japrescott:
I checked the migration guide and the docs but I wasn't able to find a way to enable client minification like  in <1.0. Does such a function exist? Or should this (and gzip'ing etc.) be handled by a reverseproxy like nginx?

PR #2154, opened 2015-06-15
Serve minified and gzipped source at server/socket.io/socket.io.min.js

boblauer:
Currently, the server only serves the unminified, uncompressed source.  This pull adds functionality to serve the minified, gzipped source at `/socket.io.min.js`.


nkzawa:
Unfortunately, this wouldn't be supported. Please see https://github.com/socketio/socket.io/issues/1847#issuecomment-61384590 for the reasons.

PR #2171, opened 2015-07-06
Actually prevent double callbacks...

KevinMartin:
We declare `sent = false` and even check `if (sent)` is truthy, but we never set it to truthy, therefore it never really *prevents double callbacks*...


rauchg:
😅Thanks!

PR #2218, opened 2015-08-23
Clarified documentation of Socket.in

futbalguy:
The documentation was a little confusing because it showed the socket.to() example under the socket.in() heading. I have made minor changes to be more clear.


nkzawa:
thx!

Issue #2199, opened 2015-08-04
Latency/Load/Overhead caused by integer indexed arrays indexOf/splice/shift etc


selay:
Recently I was debugging to find the reason for frequent server crashes, delays in transmission and server capacity issues. 

I have found several issues in socket.io and hope it might be useful to report here. Well, I had to hack it to solve the issues and it provided unbelievable improvement. However, I have only changed the parts I use and that means my current changes are not mergable and not compatible with existing modules. 

Problem: 
Room etc uses integer based arrays to keep sockets and then uses indexOf multiple times to find the index of socket. IndexOf is used many times during a single emit process. Additionally splice is used to remove a socket, and that means it re-arranges multiple big arrays for each socket changes. 
I did a small simulation with 10000 sockets connected and randomly leaving, emitting, connecting etc. 
(Compared socket.io against my hacked version which uses "associative array" where index and value are the same. No need for indexOf and splice (because no need to rearrange array indexes).)

When 10000 sockets are connected, emit is 2040 times!!! faster in the hacked version than current socket.io implementation. The server load is at least twice low, and memory usage is 30% less. 
The latency and delay issues disappeared. 
(it apparently solved another unrelated issue - when server is busy to accept a new handshake quickly, client side continues to make new polling requests and as a result, multiple connections were happening with the same socket.id. When a message comes, it was getting echoed/duplicated many times.)

I dont understand what the reason is to use integer indexed array and then search each time. As location is not known, each searches the entire array (10000 sockets multiple times for a single emit and how about 1000 sockets emit and this overhead is multiplied by 1000, a single disconnect rearranges all 10000 sockets etc * 1000). 

Actually, you can run a simple test to compare [] acess against indexOf and can see the performance difference which is over 3000 times. 

So, the current socket.io implementation is only suitable for a small-scale project where you don't expect more than 3000 sockets connecting. Or you will need unnecessarily have load balancing, multiple nodes etc.

nuclearace:
Interesting. If this is really a problem I think it would be worth seeing some improvements.

nuclearace:
@rauchg What are your thoughts?

fourpixels:
@selay - hey this sounds pretty interesting. Can you share the modifications you've made or at least any ideas what you've done. I'm pretty interested in this fix, as I've already browsed the code and thought it was quite a strange approach.

Do you simply use object with socket.id as a key? That's what I've been doing but not on that low level..

[...]

PR #2239, opened 2015-09-11
converted arrays to objects

darrachequesne:
See #2199

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -347,9 +347,9 @@ Server.prototype.of = function(name, fn){
  */
 
 Server.prototype.close = function(){
-  this.nsps['/'].sockets.forEach(function(socket){
-    socket.onclose();
-  });
+  for (var id in this.nsps['/'].sockets) {
+    this.nsps['/'].sockets[id].onclose();
+  }
 
   this.engine.close();
 
--- a/lib/namespace.js
+++ b/lib/namespace.js
@@ -51,7 +51,7 @@ var emit = Emitter.prototype.emit;
 function Namespace(server, name){
   this.name = name;
   this.server = server;
-  this.sockets = [];
+  this.sockets = {};
   this.connected = {};
   this.fns = [];
   this.ids = 0;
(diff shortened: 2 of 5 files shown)


rauchg:
Wouldn't it be better to use a `Map` here?

nkzawa:
Object would be better for now, since there is no `Map` on Node.js 0.10.

darrachequesne:
@nkzawa yes, that's my opinion too :+1: 

If we'd want to use  ̀Map` (and more generally the ES2015 features) and still keep support of node < 0.12, I think a migration to Babel (or other) would be necessary (but that's out of the scope here)

rauchg:
In the meantime we can merge this.
Moving forward, I think we should be using Babel and `babel-runtime` + runtime transformer to provide `Map`.

rauchg:
Please rebase. Current status: image

darrachequesne:
Done! :angel:

lib/index.js
```diff
@@ -347,9 +347,9 @@ Server.prototype.of = function(name, fn){
  */
 
 Server.prototype.close = function(){
-  this.nsps['/'].sockets.forEach(function(socket){
-    socket.onclose();
-  });
+  for (var id in this.nsps['/'].sockets) {
```
rauchg: You're not checking for `hasOwnProperty` here. Extensions fo the natives will make socket.io break here. Test needed for that scenario as well.

lib/namespace.js
```diff
@@ -187,9 +187,8 @@ Namespace.prototype.add = function(client, fn){
  */
 
 Namespace.prototype.remove = function(socket){
-  var i = this.sockets.indexOf(socket);
-  if (~i) {
-    this.sockets.splice(i, 1);
+  if (this.sockets[socket.id]) {
```
rauchg: `this.sockets.hasOwnProperty` would be a better test. 
See for inspiration: https://github.com/socketio/socket.io-adapter/pull/30

darrachequesne:
Hi! I opened a new PR to clean this ugly one: https://github.com/socketio/socket.io/pull/2322

PR #2247, opened 2015-09-15
add socket middleware feature, fix jsdoc and fix socket close bug

denghongcai:


darrachequesne:
Hi! I think this is really an interesting feature :+1:

Do you mind adding some tests?

Note: I may be wrong (if so, please ignore my comment), but I don't think the JSdoc and other merge commits are really useful here, as they may hide the real feature behind this PR.

denghongcai:
seems that this repo is dead, glad to know you like it :smile: 

I'll add some test

denghongcai:
#2356 @darrachequesne 
:smile:

PR #2248, opened 2015-09-16
Make it compile under Node 4.0.0; bump `socket.io-client`

pmq20:
The previous reference `automattic/socket.io-client#bf98153` points to 0.7.x of `ws` package which could not compile under Node 4.0.0.

We have to point to `ws` 0.8.0 to make it compile.


pmq20:
cf. https://github.com/websockets/ws

nkzawa:
Should be solved since it has reference of `automattic/socket.io-client#master` now.

PR #2262, opened 2015-09-28
revert 3a259d420cc32ab7e664e2e32d0d86eb86420d0e

denghongcai:
original commit will cause memory leak when middleware request is done and client is closed, `process.nextTick` will make run callback queue after client close event handler


denghongcai:
refer #2247

PR #2271, opened 2015-10-06
Working travis build status

dazorni:
Remove broken build status caused by the name change of the github organization.


rauchg:
Thank you!

PR #2287, opened 2015-10-27
fix the text alignment in firefox

championswimmer:
This is just plain strange, but using fotn size above 150% makes the alignment got crazy on firefox. No idea at all why and how. 

Check http://imgur.com/TRdhV8Y to have an idea of what I am talking about.

### DIFF
--- a/examples/chat/public/style.css
+++ b/examples/chat/public/style.css
@@ -95,7 +95,7 @@ ul {
 /* Font */
 
 .messages {
-  font-size: 150%;
+  font-size: 100%;
 }
 
 .inputMessage {


nkzawa:
Umm, don't know why but I can't reproduce the issue. Firefox on OSX.

championswimmer:
I can still reproduce this. 

Firefox 41.0.2, Linux x64

championswimmer:
I can't fathom, at all, why though

nkzawa:
Fixed via https://github.com/socketio/socket.io/pull/2477. Thanks!

Issue #2311, opened 2015-11-18
socket.join() sometimes (!) does not work


helios1138:
I have a system in which a lot of clients are constantly joining and leaving rooms and I see this weird behavior in which I do:
```js
socket.join(someRoom);
```
and them on the next line I have:
```js
console.log(io.of('/').adapter.rooms[someRoom][socket.id])
```
which prints `true` **most of the time**, but sometimes I get `undefined`

in `Socket` class source code I see this line that stops socket from joining the room:
```js
if (~this.rooms.indexOf(room)) return this;
```
So I now I see that problem is that `adapter.rooms` and `socket.rooms` are out of sync for some reason.

Also in the `Adapter.prototype.add()` I see this:
```js
if (fn) process.nextTick(fn.bind(null, null));
```

So apparently `adapter.rooms` and `socket.rooms` are updated in different event loops. Might this be causing the problem?

fourpixels:
I was also checking this today. I have no idea why the `process.nextTick` is used. Any clues?
Anyways, your pull request seems legit to me.
I still don't understand why those rooms are so complicated. There is another issue suggesting simple Object mapping, which is pretty fast compared to Array search / modifications. Might wanna check it.

darrachequesne:
Hi! This issue should be solved now, with the following merged PR:
- https://github.com/socketio/socket.io/pull/2239 (object mapping)
- https://github.com/socketio/socket.io/pull/2325 (trigger callback on already joined room)

(should be released with the forthcoming version)

PR #2312, opened 2015-11-18
fix room joining

helios1138:
fixes socketio/socket.io#2311

### DIFF
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -223,11 +223,12 @@ Socket.prototype.packet = function(packet, opts){
 Socket.prototype.join = function(room, fn){
   debug('joining room %s', room);
   var self = this;
-  if (~this.rooms.indexOf(room)) return this;
   this.adapter.add(this.id, room, function(err){
     if (err) return fn && fn(err);
     debug('joined room %s', room);
-    self.rooms.push(room);
+    if(!~self.rooms.indexOf(room)){
+      self.rooms.push(room);
+    }
     fn && fn(null);
   });
   return this;


nkzawa:
Recently, the type of `Socket#rooms` was changed to `Object`. I think this change would solve this issue too. Can you check that if possible?

https://github.com/socketio/socket.io/blob/master/lib/socket.js#L229

darrachequesne:
Hi! Please correct me if I didn't understand right, but why aren't you using the callback here `Socket.prototype.join = function(room, fn)`? 

```javascript
socket.join(someRoom, function() {
  console.log(io.of('/').adapter.rooms[someRoom][socket.id]); // should always work, right?
});
```

(since `Adapter.prototype.add()` is async)

helios1138:
@nkzawa, it does seems like *my problem* is solved, however the underlining issue is still present (even though it does not affect my end result anymore). I think potentially it can be a source of further problems.

What that means is that in the case I'm describing you will have:
```js
Socket.prototype.join = function(room, fn){
  debug('joining room %s', room);
  var self = this;
  // so *this.rooms* does not contain the *room* here
  if (this.rooms.hasOwnProperty(room)) return this;
  this.adapter.add(this.id, room, function(err){
    if (err) return fn && fn(err);
    debug('joined room %s', room);
    // but here *self.rooms* already contain the *room* sometimes
    self.rooms[room] = room;
    fn && fn(null);
  });
  return this;
};
```
Basically there is a race condition. Ideally there should be *single source of truth* regarding socket room occupancy. So the easiest way is to either make `socket.rooms` to be just a **getter** that will access `adapter.rooms` or vice versa. If for some reason it is needed to keep two separate lists (both of which can be changed by asynchronous operations), then there should be some kind of transaction logic to keep them in sync. Simple `if` condition on just one of the lists is not enough.

@darrachequesne, that's the thing - it doesn't. Async functions should **always** call the callback at the end, either with success or with error. If you look at `Socket.prototype.join()` method code, you will see that in the condition check that happens before the callback handling if the condition is `true`, it does `return this` **without calling the callback at all**. So in your code:
```js
socket.join(someRoom, function() {
  // this callback will never be actually called in the case I'm describing
  console.log(io.of('/').adapter.rooms[someRoom][socket.id]);
});
```

darrachequesne:
My understanding is that the adapter (be it the default one, redis-adapter or any other that implements the given interface) has that *truth*. As far as I understand, the local `socket.rooms` only prevents from a superfluous round-trip to the adapter.

Now, I think you're right with the callback thing, maybe this would be better?
```javascript
if (this.rooms.hasOwnProperty(room)) {
  fn && fn(null);
  return this;
}
```

darrachequesne:
Hi! I took the liberty to create a new PR https://github.com/socketio/socket.io/pull/2325 so that this fix could be included in the forthcoming release (which should happen soon!)

nkzawa:
https://github.com/socketio/socket.io/pull/2325 should solve the issue, thanks @darrachequesne !

PR #2337, opened 2015-12-07
Promise middleware

gausie:
With these changes socket.io middleware functions can return Promises instead of using the `done` callback. This does not require including a Promise polyfill in socket.io.

### DIFF
--- a/lib/namespace.js
+++ b/lib/namespace.js
@@ -112,17 +112,30 @@ Namespace.prototype.run = function(socket, fn){
   var fns = this.fns.slice(0);
   if (!fns.length) return fn(null);
 
+  function runSuccess(i) {
+    // if no middleware left, summon callback
+    if (!fns[i + 1]) return fn(null);
+
+    // go on to next
+    run(i + 1);
+  }
+
   function run(i){
-    fns[i](socket, function(err){
+
+    var result = fns[i](socket, function(err){
       // upon error, short-circuit
       if (err) return fn(err);
 
-      // if no middleware left, summon callback
-      if (!fns[i + 1]) return fn(null);
-
-      // go on to next
-      run(i + 1);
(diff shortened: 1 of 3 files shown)


fourpixels:
I'm voting with two hands **against** this pull! Because sockets are used for real-time communication, it must be as fast as possible. And promises are using *try-catch*, even in ES6, which is tremendously slow. The only benefit, in this scenario, is the syntax "sugar", which is not needed if developers works properly.

So gain eye candy in exchange for losing speed seems bad for me.

gausie:
Its completely optional for the developer though.

fourpixels:
Don't get me wrong - I just say I don't see any kind of benefits with this complication. And because it's a middleware, I still cannot see how you can chain it using a Promise?

gausie:
It's not about chaining post middleware, rather simplifying the process if your middleware function uses promises, so you just need to return the Promise instead of add a done callback. Just seems like an easy-to-implement and neat feature for socket.io to have!

capaj:
@fourpixels ever heard of [koa](http://koajs.com/)? these guys are chaining promises without any issues

fourpixels:
@capaj I guess you're mistaking Promise with generator functions. I don't see any promises in Koa at all.
But just imagine you're doing everything with Promises - a simple socket package might invoke 10 or even more try-catches. Does it sound good to you as a matter of performance? Keep in mind this is a low-level socket library which aims for speed and performance. Error handling is not the most important thing out there..

MaffooBristol:
Yeah, you can just promisify the functions yourself. I don't think they should be in a low-level lib either.

perrin4869:
This could also probably be implemented as a middleware that monkey-patches `io.use`. I may try to add it to `socket.io-as-promised` if I get time

PR #2357, opened 2015-12-25
fix bug when socket closed, socket will not remove from io.sockets.connected

denghongcai:
which will cause memory leak

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -221,20 +221,23 @@ Client.prototype.onerror = function(err){
  */
 
 Client.prototype.onclose = function(reason){
-  debug('client close with reason %s', reason);
+  var self = this;
+  process.nextTick(function() {
+    debug('client close with reason %s', reason);
 
-  // ignore a potential subsequent `close` event
-  this.destroy();
+    // ignore a potential subsequent `close` event
+    self.destroy();
 
-  // `nsps` and `sockets` are cleaned up seamlessly
-  for (var id in this.sockets) {
-    if (this.sockets.hasOwnProperty(id)) {
-      this.sockets[id].onclose(reason);
+    // `nsps` and `sockets` are cleaned up seamlessly
+    for (var id in self.sockets) {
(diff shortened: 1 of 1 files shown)


rauchg:
Can you elaborate on what causes `process.nextTick` to fix this?

denghongcai:
this problem happened on my production situation.

if client connect to server and immediately fire a transport error.
There are two functions which may run at the same tick.
```
Namespace.prototype.add
```
and
```
Client.prototype.onclose
```

the callback within  `Namespace.prototype.add` will actually run at `nextTick`

but`the callback within `Client.prototype.onclose` will run at current tick.

This behavior will let `socket` info stay in `connected` array, though it is fine because it can not be reached using normal function. On the busy situation, many clients will fire transport error, which will lead to a slightly(may be serious) memory leak.

fourpixels:
Why are these things happening on `nextTick`, is there any reason?

denghongcai:
@fourpixels It's about the way socket.io implemented.

fourpixels:
Yeah, I know, I'm asking if anyone has a clue why? :) I don't see any benefits, rather than complicating the code. Maybe it's because if there is a different adapter than the memory one, the functions will take time.. So here it's using nextTick to simulate asynchronous call?

rauchg:
@fourpixels that's exactly why. Adapters can be async.

rauchg:
@denghongcai wouldn't this be false though?
https://github.com/socketio/socket.io/blob/master/lib/namespace.js#L159

rauchg:
Assuming, like you say, the client closes and then on the next tick it gets added.

denghongcai:
https://github.com/denghongcai/socket.io/blob/socket-middleware/lib/client.js#L221

sorry, my fix is based on old code...the new code can not reproduce the problem... and the problem occured three month ago, i have to reconstruct the  test environment. Once I reproduce the problem I will reopen this pr.

PR #2395, opened 2016-01-18
Update main.js

embersdev:
applying .text() to text(input) doesnt strip html.  However, if you do .html(input).text() it does.


nkzawa:
Thx, but it seems we can just remove `cleanInput()` then? It seems data is html-escaped anyway.

darrachequesne:
Closed due to inactivity. Please reopen if needed.

PR #2410, opened 2016-01-27
reads socket client source only when its served

heldr:


nkzawa:
Thanks. but `read()` is a synchronous method, thus this could be performance problem.
Additionally, we'd like to read the source only once, even if `serveClient` was called several times.

heldr:
You're right!

palavrov:
Current implementation makes troubles with `broserify` used in `nexe` - see jaredallard/nexe#219 and jaredallard/nexe#213
I.e. it must not read the client source when it is not served. Any ideas?

lib/index.js
```diff
@@ -94,6 +88,14 @@ Server.prototype.checkRequest = function(req, fn) {
 Server.prototype.serveClient = function(v){
   if (!arguments.length) return this._serveClient;
   this._serveClient = v;
+
+  /**
+   * Socket.IO client source.
+   */
+  if (v) {
```
palavrov: To read the file just once per instatiation add `&& !this._clientSource` i.e.
```JS
if (v && !this._clientSource)
```

Issue #2152, opened 2015-06-12
Chat example -- would anyone please update it to use the newest express version?


agamemnus:
Thanks...

aHumanBeing:
Would need this as well.  Express generator now creates the http server in /bin/www file and not app.js.  Furthermore, I need the ability to maintain the socket between routes.  How may I accomplish this in a clean way?  Thank you.

darrachequesne:
That issue was closed automatically. Please check if your issue is fixed with the latest release, and reopen if needed (with a [fiddle](https://github.com/darrachequesne/socket.io-fiddle) reproducing the issue if possible).

PR #2416, opened 2016-01-30
updated express version to express 4

paradite:
updated the express version, addressing #2152

the example works fine with new version express

### DIFF
--- a/examples/chat/package.json
+++ b/examples/chat/package.json
@@ -7,6 +7,6 @@
   "private": true,
   "license": "BSD",
   "dependencies": {
-    "express": "^4.13.4"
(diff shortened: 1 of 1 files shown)


examples/chat/package.json
```diff
@@ -7,6 +7,6 @@
   "private": true,
   "license": "BSD",
   "dependencies": {
-    "express": "3.4.8"
+    "express": "^4.13.4"
```
nkzawa: we don't use `^` for specifying dependencies.

paradite:
@nkzawa fixed. Thanks for the note!

PR #2432, opened 2016-02-05
Few optimizations

marijaselakovic:
I suggest to replace for in loop with hasOwnProperty check with significantly cheaper Object.keys and for loop (jsperf: https://jsperf.com/object-keys-vs-for-in-perf). In String.prototype.of method you call String constructor  which usually receives some string as an input. It is cheaper to use implicit string conversion, and performance benefit is even larger for inputs of type string (jsperf: http://jsperf.com/implicitstringconversion)


darrachequesne:
`nsps` hash will be replaced by a `Map` starting 2.x. Thanks though!

PR #2456, opened 2016-02-22
Update License year to 2016

imkimchi:


darrachequesne:
Let's be foolish! Thanks.

PR #2477, opened 2016-03-09
Fix chat display in Firefox

callahad:
Not sure why `.username` was floated, but it causes the messages to steadily march to the right in Firefox, per screenshot:

socket

Removing the float (or setting `.messages > * { clear: both; }`) results in identical, correct display in both Chrome and Firefox.

### DIFF
--- a/examples/chat/public/style.css
+++ b/examples/chat/public/style.css
@@ -128,7 +128,6 @@ ul {
 }
 
 .username {
-  float: left;
   font-weight: 700;
   overflow: hidden;
   padding-right: 15px;


callahad:
ping @rauchg

nkzawa:
Thanks. it's weird but this happens only on Firefox for Windows, not on Firefox for Mac.

callahad:
Odd, I *was* seeing it on Mac (note the scrollbar in the screenshot), but hey, not it should woek everywhere for everyone. Thanks for merging! :)

PR #2482, opened 2016-03-10
serve sourcemap for socket.io-client

paradite:
This fix allows server to serve the sourcemap of socket.io-client generated by webpack builder using `gulp build` task.

The corresponding PR for `socket.io-client` is https://github.com/socketio/socket.io-client/pull/953

Tested in Chrome and Firefox using a custom package of `socket.io-client` with sourcemap file. It also catches the error when there is not sourcemap available in `socket.io-client`.

Screenshots of how the generated source map looks in browsers:
screen shot 2016-03-10 at 4 00 40 pm
screen shot 2016-03-10 at 4 00 29 pm

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -25,7 +25,13 @@ module.exports = Server;
  */
 
 var clientSource = read(require.resolve('socket.io-client/socket.io.js'), 'utf-8');
-var clientSourceMap = read(require.resolve('socket.io-client/socket.io.js.map'), 'utf-8');
+var clientSourceMap = '';
+try {
+  clientSourceMap = read(require.resolve('socket.io-client/socket.io.js.map'), 'utf-8');
+} catch(err) {
+  debug('could not load sourcemap file');
+}
+
 
 /**
  * Server constructor.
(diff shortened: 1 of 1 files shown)


darrachequesne:
@paradite hi! could you rebase after https://github.com/socketio/socket.io/pull/2603 and https://github.com/socketio/socket.io/pull/2567 please?

lib/index.js
```diff
@@ -290,6 +300,32 @@ Server.prototype.serve = function(req, res){
...
+    if (clientVersion == etag) {
+      debug('serve client 304');
+      res.writeHead(304);
+      res.end();
+      return;
+    }
+  }
+
+  debug('serve client sourcemap');
+  res.setHeader('Content-Type', 'application/javascript');
```
darrachequesne: Also, according to [this](http://stackoverflow.com/questions/19911929/what-mime-type-should-i-use-for-javascript-source-map-files), it seems the content-type should rather be `application/json`.

paradite:
sure. i will check what to do.

darrachequesne:
@paradite thanks a lot, sorry for the delay (...)

paradite:
Ah, sorry I did merge instead of rebase, is that okay? 

I've updated the code based on  #2603 and #2567 and I retained the try catch block in case of inconsistent versions between server and client resulting in no sourcemap file.

darrachequesne:
@paradite thanks!

PR #2486, opened 2016-03-13
Add ESLint

digawp:
Brief:
- Add ESLint
- Fix code to make it conform to the standard
- Add gulp task for linting (`gulp lint`)
- Run lint before test

Important changes/refactoring due to ESLint will be annotated on the diff


lib/index.js
```diff
@@ -194,27 +193,22 @@ Server.prototype.origins = function(v){
...
-    srv = Number(srv);
-  }
-
-  if ('number' == typeof srv) {
+  // handle port as a string and number
+  if (!isNaN(srv)) {
     debug('creating http server and binding to %d', srv);
-    var port = srv;
-    srv = http.Server(function(req, res){
+    var port = Number(srv);
```
digawp: Somewhat major refactoring here.

To avoid the use of `==`, I modified the handling of port number given as `string`; it is now handled together with port number given as `number` by checking if `srv` is not `NaN` (i.e. `!isNaN(srv)` on line 204)

Passed the test, but let me know if there is an issue with this

lib/namespace.js
```diff
@@ -62,14 +63,14 @@ function Namespace(server, name){
  * Inherits from `EventEmitter`.
  */
 
-Namespace.prototype.__proto__ = Emitter.prototype;
+util.inherits(Namespace, Emitter);
```
digawp: Modified inheritance to use node.js `util`. This is because ESLint throws an error, mentioning that the `__proto__` way is deprecated.

darrachequesne:

PR #2492, opened 2016-03-16
Update package.json

westy92:
I updated the repository URL and removed an unnecessary path to gulp.

Even if gulp is not installed globally, the script 'test' will still run fine.  "...executables will be added to the PATH for executing the scripts." (https://docs.npmjs.com/misc/scripts#path)


nkzawa:
thx!

PR #2496, opened 2016-03-19
Added test for compat version transformed by babel

paradite:
This allows to test both the **current** source files in `lib` directory and **compat** version transformed by babel in `dist` directory by setting the environmental variable `MODE`.
In the future when we move the code base to es6, we can make sure both versions pass the test.

Major changes:
- explicit `require` path (`../lib` instead of `..`) in the test, allowing us to switch the default version we serve inside `package.json`'s main between `lib`(current version) and `dist`(compat version) and not affect the tests.
- more verbose output for `mocha` reporter to check if we are testing compat or current version, it prints `testing compat version` in the console (I can change it back to dot since I have tested this in CI and it is running the correct versions)
- build the compat version to `dist` before testing.
- set the environmental variables to instruct the CI to run both current test and compat tests for node 0.10, 0.12, 4. In the future we could drop the current test for node 0.10 and 0.12 when we move to es6.


rauchg:
`MODE` is not very explicit as an env var. We should also probably document this test behavior in the README. Thank you!

paradite:
@rauchg 
I changed the env var name to `TEST_VERSION ` (maybe you have better suggestions? `VERSION_TO_TEST` is an alternative)

I also added a `TESTING` section in README.md
https://github.com/nus-fboa2016-si/socket.io/tree/compat-test

paradite:
@rauchg This PR is needed for the next PR on es6 refactoring which I am working on. 
I have refactored all the anonymous functions into arrow functions and the babel transformed version is working well.
I hope you can review this PR soon so that I can send my PR for es6 refactoring.

rauchg:
Awesome stuf @paradite

PR #2502, opened 2016-03-29
updated Express instructions so they work.

Pomax:
The listed instructions do not work, this update does. I spent a good 2 hours trying to figure out why the heck nothing worked until I ran across [this SO post](http://stackoverflow.com/questions/24222483/socket-io-1-0-express-4-2-no-socket-connection/24222869#24222869) that lists the code change in this PR, which worked perfectly with modern Express.

Another update to the site's http://socket.io/docs/ will be necessary, as that also lists code for express 3/4 that simply does not work.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -40,10 +40,9 @@ function.
 
 ```js
 var app = require('express')();
-var server = require('http').createServer(app);
+var server = app.listen(3000);
 var io = require('socket.io')(server);
 io.on('connection', function(){ /* … */ });
-server.listen(3000);
 ```
 
 ### In conjunction with Koa


nkzawa:
As far as I tried with the latest `express`, it works fine for me.

Pomax:
that's quite surprising, as the code as-is does not start the express app, but a different, standard nodejs `http` server. This worked for you with a straight up copy-paste and express 4?

nkzawa:
Not exactly copy-paste, but `equire('http').createServer(app);` works fine.

Pomax:
What did you have to change? The docs are to illustrate the "how to use this", so if any of it is not a straight copy paste, that's worth fixing (even if that fix differs from the code currently in this PR =)

nkzawa:
I just added some routing for express and socket.io, so I think the doc is ok. I wonder why it's not for you. Can you confirm again if possible?

baconface:
I don't know why the original example doesn't work for you. It works for me with Node 5.X and Express 4.X.

baconface:
What are your Node and Express versions?

Pomax:
Retried in a clean dir with a fresh `npm install express` and `npm install socket.io`, using node 5.10.1 in which the doc code does not work, but for different reasons than this PR tried to fix.

Code that worked:
```
var app = require('express')();
var server = require('http').Server(app);
var io = require('socket.io')(server);

server.listen(80);

app.get('/', function (req, res) {
  // note casing, "res.sendfile" throws deprecation warnings
  res.sendFile(__dirname + '/index.html');
});

// fairly crucial, not currently mentioned in the docs:
app.get('/socket.io/socket.io.js', function (req, res) {
  res.sendFile(__dirname + 'node_modules/socket.io-client/socket.io.js');
});

io.on('connection', function (socket) {
  socket.emit('news', { hello: 'world' });
  socket.on('my other event', function (data) {
    console.log(data);
  });
});
```

Issue #2491, opened 2016-03-16
Socket.io prepends '/#' to server ids


FabioAntunes:
I'm having a weird issue, I have just cloned a repo that was working perfectly fine, but for no aparent reason it isn't anymore.

Basically what's happening is that my socket.io server instance is prepending '/#' to the ids making them different from the client side.

So on the server side I have **/#zG3wncZxZvgRE7LKAAAA** and on the client side I have **zG3wncZxZvgRE7LKAAAA**

Also when the client is receiving a message, it's possible to see on the frames tab this:

    4undefined

This is my app.js and `config.room = fabio1`:

```javascript
    import config from './config';
    import io from './factories/io';
    
    io.on('connection', (socket) => {
      socket.join('dmClient');
      socket.join(config.room);
      console.log('socketId', socket.id);
    });
```

And this is my **/factories/io.js**
```javascript
    import socketIo from 'socket.io';
    const io = socketIo(3000);
    
    export default io;
```
More info:

 - socket.io - 1.4.5
 - node - v4.4.0

creole:
See https://github.com/socketio/socket.io/issues/2475 and https://github.com/socketio/socket.io/issues/2405

Issue #2475, opened 2016-03-08
Default room name is not socket id


tw0517tw:
I'm using socket.io 1.4.5.

if I do
```
io.to(socketId).emit('some msg');
```
the message will not be emitted.

after digging a little bit I found that I should prepend `/#` before socke id.
```
socketRoomName = '/#' + socketId;
io.to(socketRoomName).emit('some msg');
```

baconface:
A socket ID is not the same as a room ID. The socket ID is unique per client. So the above is sending 'some msg' to that specific client and not a room.

Think of a room as a group of people and a socket as a singular person in this context.

baconface:
Also ```io.emit('some msg');``` will send it to everyone in the default room. So if you have no use for multiple rooms you can omit it's usage completely from your code and let socket.io handle it.

tw0517tw:
but the document says that every socket joins a room of its socket id.
http://socket.io/docs/rooms-and-namespaces/#default-room

I think this is related to #2405 , that client and server returns different format of socket id.

hocuspucus:
You're right. Just ignore the guy above.

I lost two hours trying to figure out what was going on. Turns out when I did THISID = socket.id it got stored as "EgdohcmzQZnc_FzHAAAA" instead of ('/#' + "EgdohcmzQZnc_FzHAAAA") which is the true socket id. So server side socket.id is /#EgdohcmzQZnc_FzHAAAA, but client side socket.id is EgdohcmzQZnc_FzHAAAA. Who is responsible for this?

tw0517tw:
Dig a little bit in the source I found that it's this line
https://github.com/socketio/socket.io/blob/master/lib/socket.js#L62
form this commit
https://github.com/socketio/socket.io/commit/b73d9bea4efb48277eee685763026ff2df5a79ab#comments
break these thins

tw0517tw:
PR merged.

Issue #2451, opened 2016-02-18
update 1.3.7 to 1.4.5 stop sending messages to roms


diegoug:
Hello, I have a problem with socket.io update version 1.3.7 to 1.4.5 and stop sending messages to the rooms, you know what might have happened?

My code is very simple, it is only that:

```
// socket to room, using roomKey
socket.join(roomKey);
// room emmit
this.io.to(roomKey).emit(channel, message);
```

But when I want to send a message by a room in particular nothing happens, and if I go back to version 1.3.7 works again, you know it can be?

CharGomez:
Same happens to me, any clue ? Thx.

svarunan:
I too have this issue after new update. 
on server side socket.id gives me /#LF6L4-bVXD4u3fT9AAAA
on client side its LF6L4-bVXD4u3fT9AAAA
'/#' creates all new problem now. 
only io.emit works bit io.sockets.connected returns ["/#LpjTrYgEzpl2wbi2AAAB"]

Team, Please help us get a workaround or give a solution to this.

oliversalzburg:
We had a similar issue that was primarily caused by this change: https://github.com/socketio/socket.io/issues/1956

jcsla:
I had same issue.. plz help.

dottodot:
Same here with 1.4.6, not used socket.io in a while and thought I was doing something wrong, turns out rooms don't work at all as downgrading to 1.3.7 works as documented.

darrachequesne:
That issue was closed automatically. Please check if your issue is fixed with the latest release, and reopen if needed (with a [fiddle](https://github.com/darrachequesne/socket.io-fiddle) reproducing the issue if possible).

Issue #2405, opened 2016-01-25
Server socket id doesn't match client's socket id


stgogm:
For some reason, after upgrading to 1.4.4, the sockets on the server return the `socket.id` property as `/#BA0yIQhFb5R55OnvAAAC` but in the client as `BA0yIQhFb5R55OnvAAAC`.

This inconsistency has broken one of my apps. Am I doing something wrong?

nkzawa:
That was changed on 1.4.0. Maybe we should change `socket.id` on client to the same format with server. cc @rauchg

You can use `socket.client.id` on server for now, which would be matched with client's `socket.id`.

stgogm:
Thanks, but I think this behavior breaks functionality and if you installed `socket.io` via `npm` and it's defined as `"^1.3.6"`, for example, it'll definitively break your app if you rely on `socket.id` to achieve some functionality.

One more thing, if now `socket.id` is not equal to `socket.client.id` that means that when you want to emit a message to a specific socket it'll be sent to both `socket.id` and `socket.client.id`?

smcardle:
I also have this issue.

The server side pre-pends the default namespace but the client does not.
This has caused me SERIOUS issues with our app after an upgrade. We missed 1.4.0 and went straight to 1.4.4. 
It's good to know that socket.client.id on the server side is the same as the client side id.

However, a fix for this would be good....

I also noticed a few other issues with this upgrade..

1. Counting the number of sockets in a room
My original code was a follows :

                var numClients = 0;
                var clients = sio.sockets.adapter.rooms["test"];
                if(clients)  {
                    numClients = Object.keys(clients.length;
                }
i.e. sio.sockets.adapter.rooms[data.room] returned an array containing the socket ids of the sockets that had joined the "test" room

After this upgrade sio.sockets.adapter.rooms["test"] now returns a Room object that contains an array called sockets that contains the socket ids instead, requiring a change to the code as follows

                var numClients = 0;
                var room = sio.sockets.adapter.rooms["test"];
                if(room)  {
                    numClients = Object.keys(room.sockets).length;
                }

2. socket.io-redis DOES NOT WORK with this version as broadcasting and emitting events to clients in a room does not result in any of those clients receiving a message. 
However, emitting to the self i.e. socket.emit(...) does work.

Any chance these issues will be resolved soon ???

Steve

LordMajestros:
@smcardle Socket.io-redis works for me but I had to update to 1.0.0 check your version.

[...]

PR #2506, opened 2016-04-01
stop modifying socket id on server side

tw0517tw:
should fix break since https://github.com/socketio/socket.io/commit/b73d9bea4efb48277eee685763026ff2df5a79ab
#2451  #2405  #2475  #2491
but this could be another breaking change, I'm not sure what will this affect to those who parse nsp name from server side id.


tw0517tw:
fixing tests, close for now.

Issue #2507, opened 2016-04-01
graceful shutdown test fail on windows


tw0517tw:
It seems to be a bad idea to put `process.execPath` in `exec`, because it contains space in `execPath` on Windows

```
$ npm run test

> socket.io@1.4.5 test D:\GitHub\socket.io
> gulp test

[20:41:20] Using gulpfile D:\GitHub\socket.io\gulpfile.js
[20:41:20] Starting 'test'...
  ..............................

  29 passing (9s)
  1 failing

  1) socket.io close graceful close should stop socket and timers:
     Error: Command failed: C:\Program Files\nodejs\node.exe D:\GitHub\socket.io
\test\fixtures\server-close.js
'C:\Program' ���O�����Υ~���R�O�B�i���檺�{���Χ妸�ɡC

      at ChildProcess.exithandler (child_process.js:202:12)
      at maybeClose (internal/child_process.js:850:16)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:215:5)
```

darrachequesne:
That issue was closed automatically. Please check if your issue is fixed with the latest release, and reopen if needed (with a [fiddle](https://github.com/darrachequesne/socket.io-fiddle) reproducing the issue if possible).

PR #2508, opened 2016-04-01
add quote to exec paths, prevent error when spaces in path

tw0517tw:
#2507


tw0517tw:
But my test on Windows is still failed due to 2000ms time out in same test case

nkzawa:
Hi, thanks for your PR!. Can you fix the conflict?

tw0517tw:
Hi @nkzawa , thanks for reviewing. 
I've solved the conflict.

PR #2509, opened 2016-04-01
stop prepend /# before id when no namespace

tw0517tw:
#2451 #2405 #2475 #2491 
This should fix break since b73d9be and #2239 

Since I'm not using namespaces, I need someone else help to check if this would break them.

### DIFF
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -59,7 +59,7 @@ function Socket(nsp, client){
   this.nsp = nsp;
   this.server = nsp.server;
   this.adapter = this.nsp.adapter;
-  this.id = nsp.name + '#' + client.id;
+  this.id = nsp.name !== '/' ? nsp.name + '#' + client.id : client.id;
   this.client = client;
   this.conn = client.conn;
   this.rooms = {};


darrachequesne:
Shouldn't this be `nsp.name != '/' ?...` ?

Else, this might actually be a good idea, since most users actually use only one namespace (the default '/').

tw0517tw:
Yes, it should be `nsp.name !== '/' ?...`.
Fixed.

nkzawa:
Thanks for your PR.

We are going to change `socket.id` on client to the same format with server, which should solve the root cause of the problem.
Let me know if you have any ideas about that.

tw0517tw:
It's good (and important?) to have same format of `socket.id` on both side (again). Hope to see that release soon.

tgabi333:
As i see this change is not included in current (v1.4.8) release, can somebody help me out why?

jdc20181:
Sometimes things cause issues- and are tested for functionality for the
best results before released (this is in general for any project) so they
might not have tested the feature yet. Or may think it will hurt something
else. Stay paitent :)

PR #2514, opened 2016-04-05
Es6 arrow functions for anonymous functions

paradite:
**node v4 LTS does not have full support for the [rest parameters](http://node.green/), whereas node v6 does, I am waiting for node v6 to become more popular to continue working on this PR**

This PR refactors existing codebase's anonymous functions into es6 arrow functions.
A new `dist` folder holds the babel transpiled es5 code which is served by default by using `"main": "./dist/index",` in `package.json` (there is no `index.js` for this repo so `package.json` defines the entry point)

Due to the new Lexical `arguments` (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions)
I have changed one use of `var args = Array.prototype.slice.call(arguments);` into rest parameters syntax `...args`. 
see: https://github.com/nus-fboa2016-si/socket.io/blob/ae6379f8d60b275e9fc2bbb2de6639e11a3b180d/lib/socket.js#L346

However, the rest parameters feature is behind the staging flag for node.js:
https://nodejs.org/en/docs/es6/#which-features-are-behind-the-es_staging-flag
So the test for current code will fail on node due to syntax error, one way is to test using the `harmony` flag: 

```
node --harmony `which gulp` test && node --harmony `which gulp` test-compat
```

For the sake of simplicity, I have removed the tests for the current code and only test the compat version of the code transpiled by `babel`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,10 @@ sudo: false
 before_install:
   - npm install -g npm@'>=1.4.3'
 language: node_js
-node_js:
-  - "0.10"
-  - "0.12"
-  - "4"
+# node_js:
+#   - "0.10"
+#   - "0.12"
+#   - "4"
 
 git:
   depth: 1
--- a/dist/client.js
+++ b/dist/client.js
@@ -1,0 +1,256 @@
+'use strict';
+
+/**
+ * Module dependencies.
+ */
+
+var parser = require('socket.io-parser');
+var debug = require('debug')('socket.io:client');
+
+/**
+ * Module exports.
+ */
+
+module.exports = Client;
+
+/**
+ * Client constructor.
+ *
+ * @param {Server} server instance
+ * @param {Socket} conn
+ * @api private
+ */
+
+function Client(server, conn) {
+  this.server = server;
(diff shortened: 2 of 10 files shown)


amit3vr:
This would be bad news for developers who aren't using the latest node.js version/don't support ES6 syntax. :cry: 

This PR might make the code much readable, but your'e risking compatibility issues.

Good job anyways.

_btw i'm looking forward for your response to my comment. i'm really excited to here more opinions about this issue, and hopefully start a discussion around this subject._

paradite:
Yup, I think I will wait until node v6 to become more mainstream before updating this PR.

amit3vr:
Haha. I don't think it's all about mainstream-ness (is that a word?) - but it's more about taking care of as much edge cases as possible.
This is why the trade-of of compatibility for readability (with no additional benefits) seems unnecessary.

darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/3d760b71d7ee480dfab4b9f336a5d90d970c6b3f (finally). Thanks!

PR #2525, opened 2016-04-16
Update .gitignore

ShavidzeT:


darrachequesne:
I don't think that belongs here, please reopen if needed!

PR #2532, opened 2016-04-30
Document that constructor options are indeed optional

mkay581:
Looks as though the `opts` argument passed to the constructor is optional, so just ensuring that it is documented as such so our IDE's won't scream at us about it. :)


darrachequesne:
Closed by https://github.com/socketio/socket.io/pull/2465. Thanks!

PR #2601, opened 2016-06-28
Server supports all event emitter methods

perrin4869:
Previously, only `on` and `emit` methods were supported by Socket.IO server. This PR iterates over all EventEmitter methods and proxies them over to the main namespace.


darrachequesne:
Thanks a lot!

PR #2603, opened 2016-06-30
Make ETag header comply with standard.

kentonv:
The standard says that an ETag must be surrounded in double quotes:

https://tools.ietf.org/html/rfc7232#section-2.3

Although browsers tend to be lenient, omitting the quotes can confuse/break some kinds of proxies and other tools that demand compliant formatting. For example, Sandstorm.io enforces strict HTTP usage for security reasons and will block responses with invalid ETags.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -272,9 +272,13 @@ Server.prototype.attachServe = function(srv){
  */
 
 Server.prototype.serve = function(req, res){
+  // Per the standard, ETags must be quoted:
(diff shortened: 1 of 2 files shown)


kentonv:
Oops, updated tests -- now fails at the same place that it fails on the master branch.

ocdtrekkie:
Apparently this is the issue that has been haunting me since February 25th of this month and I didn't know it.

darrachequesne:
Thanks!

PR #2606, opened 2016-07-04
Prevent null from being accepted as argument

ianbrode:
Null can be passed as "srv" argument and will fail on "srv.listen" check.


darrachequesne:
@ianbrode thanks!

PR #2628, opened 2016-07-31
Add a local flag that can prevent the adapter (redis) from publishing…

tal-beja:
… the emit to the pub\sub server


tal-beja:
Works with this pull request:
https://github.com/socketio/socket.io-redis/pull/119

darrachequesne:
Thanks!

PR #2661, opened 2016-08-29
Implementing classes using the new ES6 syntax

leonardo-silva:
Hi,

In my PhD research, I am using socket.io in some experiments.

As a result, I detected the existence of 4 main "classes" in the current codebase (Namespace, Client, Server, Socket).

So, I decided to rewrite this code to follow the new syntax for classes provided by ECMAScript 2016, which I think is much more readable. All changes are just syntactical. I was also very careful to preserve the code style.
Finally, all tests are green in the migrated code.

Do you think it is worth to integrate these changes to the project?

Thank you

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1,3 +1,4 @@
+'use strict';
 
 /**
  * Module dependencies.
--- a/lib/client.js
+++ b/lib/client.js
@@ -9,12 +9,6 @@ var debug = require('debug')('socket.io:client');
 var url = require('url');
 
 /**
- * Module exports.
- */
-
-module.exports = Client;
-
-/**
  * Client constructor.
  *
  * @param {Server} server instance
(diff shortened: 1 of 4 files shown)


leonardo-silva:
The build is failing exactly because I am proposing a migration to the ES6 syntax.

MaffooBristol:
Hi Leonardo,

I like it in theory, however I don't believe it is purely syntactical because it means that we drop support for legacy versions of node, which aren't actually that old in the whole scale of things, and still widely used.

If you look at your builds, [they fail on old versions](https://travis-ci.org/socketio/socket.io/jobs/155978959) because of not supporting classes, but also [fail in the newer versions](https://travis-ci.org/socketio/socket.io/jobs/155978962) because of some sort of breaking change around possibly [v1.4.6](https://travis-ci.org/socketio/socket.io/builds/127412683) (although I think that's a red herring) or more likely [v1.4.8](https://travis-ci.org/socketio/socket.io/builds/139932699).

I think it's a nice idea to move towards ES6, but maybe not something that should happen for a while on a project like this. Would require a major breaking release in order to completely deprecate older versions.

Dunno what anyone else's opinion on this is, however?

Cheers!
Matt

Disclaimer: not a maintainer. Ooh, that rhymes.

leonardo-silva:
Hi Matt,

When I checkout the original version from socketio:master I still get the same build fail [Uncaught Error: expected undefined to equal 'aa'](https://travis-ci.org/socketio/socket.io/jobs/155978962) after running the tests. Maybe I am missing something here but it seems that it is not related to the es6 syntax.

Yes, I agree with you about the dilema of moving towards ES6/ES7 and keeping compatibility with older versions.

Cheers

MaffooBristol:
Yes, sorry, maybe I wasn't 100% clear. I meant that that the error you just mentioned is nothing to do with the ES6 changes. Something was changed a few versions back that have broken the tests regardless.

darrachequesne:
Let's do this for `2.0.0`!

darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/3d760b71d7ee480dfab4b9f336a5d90d970c6b3f. Thanks!

PR #2675, opened 2016-09-10
Update the chat example

darrachequesne:
- make the example autonomous (`npm install && npm start`)
- add disconnect/reconnect logs (and send username on reconnection)

### DIFF
--- a/examples/chat/README.md
+++ b/examples/chat/README.md
@@ -6,11 +6,8 @@ A simple chat demo for socket.io
 ## How to use
 
 ```
-$ cd socket.io
-$ npm install
-$ cd examples/chat
-$ npm install
-$ node .
+$ cd socket.io/examples/chat
+$ npm install && npm start
 ```
 
 And point your browser to `http://localhost:3000`. Optionally, specify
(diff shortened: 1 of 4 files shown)


MaffooBristol:
Why make it autonomous? Is it to make compilation easier for beginner devs or some other reason?

Deffo a :+1: for the extra events though :smile:

darrachequesne:
@MaffooBristol yes, it feels easier to run it as a standalone (to me). It may be a bit superfluous though.

MaffooBristol:
Hmm, maybe. But then is it not better for the examples to be entirely tied to the current source version rather than npm's version? Also it means that anyone editing the source can see whether it breaks the chat example, as a form of testing? Just not sure if I think it's the best idea personally.

KevinHoughton:
I'm getting a 'cannot find module ../..' when trying to follow this example.

darrachequesne:
@KevinHoughton the `require('../..')` [here](https://github.com/socketio/socket.io/blob/1.7.3/examples/chat/index.js#L5) is a reference to the root repository, hence the installation steps:
```
$ cd socket.io
$ npm install
$ cd examples/chat
$ npm install
$ npm start
```

You can also replace it by `require('socket.io');` (and `npm i socket.io`, of course).

Issue #2544, opened 2016-05-11
Problem with disconnect - "client error"


JaroslavKrenar:
Hello,

There is something wrong if connection is closed with reason "client error".

Debug on server side:

**socket.io:client client close with reason client error +5s
socket.io:socket closing socket - reason client error +0ms**

and only event 'disconnect' is called on server side with reason "client error"

Debug on client side:

log is empty and none event is called on client side ("error", "disconnect" etc...). Client seems to be connected without problem, but only one way communication is working - from client to server. Server is able to receive messages from client, but client receives nothing.

We are using a latest version 1.4.6.

contrasam:
I had the same issue with v1.4.6, so I rolled back to v1.4.5 in my code, I guess thew new version has some breaking changes

borisovg:
I have also hit this issue - the client doesn't get the memo that connection is supposed to be closed.

Code to reproduce:

```javascript
var i = 0,
    port = 11111,
    server = require('socket.io')(port),
    client = require('socket.io-client')('http://localhost:' + port);

server.on('connect', function (socket) {
    console.log('SERVER :: connect');

    socket.on('error', function () {
        console.log('SERVER :: error');
    });
    
    socket.on('reconnect', function () {
        console.log('SERVER :: reconnect');
    });
    
    socket.on('disconnect', function () {
        console.log('SERVER :: disconnect');
    });

    socket.on('/ping', function (i) {
        console.log('SERVER :: ping ' + i);

        if (i > 2) {
            throw 'Spanner';
        }

        socket.emit('/pong', i);
    });
});

client.on('connect', function () {
    console.log('CLIENT :: connect');
});

client.on('error', function (err) {
    console.log('CLIENT :: error');
    console.log(err.code, 'in the works');
});

client.on('reconnect', function () {
    console.log('CLIENT :: reconnect');
});

client.on('disconnect', function () {
    console.log('CLIENT :: disconnect');
});

client.on('/pong', function (i) {
    console.log('CLIENT :: pong ' + i);
});

setInterval(function () {
    console.log('CLIENT :: socket.connected is ' + client.connected);
    client.emit('/ping', i);
    i += 1;
}, 1000);
```

Relevant log file portion:

```
[...]
CLIENT :: socket.connected is true
Fri, 16 Sep 2016 11:37:33 GMT socket.io-client:manager writing packet {"type":2,"data":["/ping",3],"options":{"compress":true},"nsp":"/"}
[...]

PR #2682, opened 2016-09-17
Avoid swallowing exceptions thrown by user event handlers

Nibbler999:
Currently these are caught by socket.io and treated as client errors. See https://github.com/socketio/socket.io/issues/2544#issuecomment-247599752


darrachequesne:
Thanks!

PR #2687, opened 2016-09-21
Nojquery

MikePap:
I created a new file "main2.js" in the directory  ../socket.io/examples/chat/pubblic
This file is alternative to "main.js". It's does not use jquery library. 
I modified ../socket.io/examples/chat/index.js by adding a variable (userNames) to show user names connected. I commit all these in the "nojquery" branch.
best regards

### DIFF
--- a/examples/chat/index.js
+++ b/examples/chat/index.js
@@ -15,6 +15,7 @@ app.use(express.static(__dirname + '/public'));
 // Chatroom 
 
 var numUsers = 0;
+var userNames = [];							
 
 io.on('connection', function (socket) {
 	var addedUser = false;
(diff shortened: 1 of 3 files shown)


MaffooBristol:
1. Why is a no-jquery version of the chat example needed?
2. What's with the changes to indentation?

MikePap:
Why using jquery today? I used jquery years ago for cross-browser problems but today browsers support are better. I don't understand the need to load a library when it doesn't need. To me reading pure javascript code is much easier.
About indentation (2), it was just for my editor. It's not a commit to give importance.
About "random commented out lines" (5) you are right. It has happened for rush.
Anyway I congratulate you for your work, best regards

MaffooBristol:
I get where you're coming from... but this is just a chat example made to teach people the fundamental basics of socket.io, and so anything that distracts the least from that is best, I reckon. The vanilla JS code added is in my opinion just a bit bloated and confusing, when functions like `$.filter` are really easy to understand what they're doing. But then the jQuery vs. no-jQuery argument is in general pretty saturated and isn't really going to go anywhere, so it's more about whether removing it would help the project and the people using it or not, and I don't really think removing it gains anything.

MikePap:
I did not want to take the discussion elsewhere. My request was not to remove your jquery script but to propose another script to put beside it as alternative.
I think that there are people that use jquery and people that don't use it. That could enhance the project, in my opinion.

MaffooBristol:
It's not my project, I just follow it and try and help out with issues and PRs where I can, but I'm certainly not a defining force in anything.

I didn't mean to be rude in anything I said, only that there are certain things that will and won't be pulled into the repo and I don't think that this is something that will be pulled because as an example piece of code, anyone can do with it as they will, and the current implementation is better for new devs than your proposition in my opinion.

MikePap:
No problem. I respect opinions. The world is beautiful because it is varied.

darrachequesne:
Thanks for the work here!

Yet I'm not sure it really simplifies the learning for newcomers, so let's close that.

PR #2691, opened 2016-09-27

bananaappletw:
There already exists client function within test file `socket.io.js` for create client.
Maybe someone didn't notice when writing test script.


bananaappletw:
Later, I could fix one test case.
`should see query parameters sent from secondary namespace connections in handshake object`

bananaappletw:
I think this test case is written by @whattokingu

bananaappletw:
Also @rauchg

PR #2722, opened 2016-10-17

bananaappletw:
There already exists client function in test script for creating client.
This commit simplify the test script.

### DIFF
--- a/test/socket.io.js
+++ b/test/socket.io.js
@@ -644,31 +644,6 @@ describe('socket.io', function(){
       });
     });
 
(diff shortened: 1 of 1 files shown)


bananaappletw:
@darrachequesne Could you take a look of this?
Thanks.

test/socket.io.js
```diff
@@ -644,31 +644,6 @@ describe('socket.io', function(){
       });
     });
 
-    it('should fire a `disconnecting` event just before leaving all rooms', function(done){
```
darrachequesne: Hi! Why would you delete that test?

PR #2731, opened 2016-10-21

bananaappletw:
I recreated the pull request.
cc: @darrachequesne

### DIFF
--- a/test/socket.io.js
+++ b/test/socket.io.js
@@ -1546,9 +1546,7 @@ describe('socket.io', function(){
       var srv = http();
       var sio = io(srv);
       srv.listen(function() {
-        var addr = srv.listen().address();
-        var url = 'ws://localhost:' + addr.port + '?key1=1&key2=2';
-        var socket = ioc(url);
+        var socket = client(srv, {query: {key1: 1, key2: 2}});
         sio.on('connection', function(s) {
           var parsed = require('url').parse(s.request.url);
           var query = require('querystring').parse(parsed.query);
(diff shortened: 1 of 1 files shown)


test/socket.io.js
```diff
@@ -1546,9 +1546,7 @@ describe('socket.io', function(){
       var srv = http();
       var sio = io(srv);
       srv.listen(function() {
-        var addr = srv.listen().address();
-        var url = 'ws://localhost:' + addr.port + '?key1=1&key2=2';
-        var socket = ioc(url);
+        var socket = client(srv, {query: {key1: 1, key2: 2}});
```
darrachequesne: Isn't the meaning of the test updated there?

bananaappletw: Nope.
Actually, this is the same meaning as the previous test.
This pull request just simplify the code.

bananaappletw: https://github.com/socketio/socket.io-client/blob/master/lib/index.js#L44
In here, the query is parsed from the url.
`key1=1&key2=2`
You could simply add debug code in socket.io-client to see the value of `opts.query`.
Eventually, they are parsed as the same value.

bananaappletw: @darrachequesne

bananaappletw: According this:
https://github.com/socketio/socket.io/blob/master/.github/PULL_REQUEST_TEMPLATE.md
[Improve code quality]

darrachequesne:
@bananaappletw thanks!!

PR #2745, opened 2016-11-08
Don't drop query variables on handshake

### LINKED ISSUE #2712: socket.io 1.5.0 backwards incompatible with socket.io-client < 1.5.0
We just had a breaking application because our server socket.io upgraded before our clients could.
#2422 seems to be the culprit as it strips all query parameters that are not `t`, `EIO` or `transport`.  With our `1.3.*` client, we need the `token` parameter to be parsed as well.

perrin4869:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
Right now parameters passed during handshake, such as tokens, are being dropped.

### New behaviour
Parameters are preserved as intended

Fixes #2712

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "debug": "2.2.0",
     "engine.io": "1.7.2",
     "has-binary": "0.1.7",
-    "object-assign": "^4.1.0",
+    "object-assign": "4.1.0",
     "socket.io-adapter": "0.4.0",
     "socket.io-client": "1.5.1",
     "socket.io-parser": "2.3.1"
--- a/package.json
+++ b/package.json
@@ -24,12 +24,13 @@
     "test": "gulp test"
   },
   "dependencies": {
+    "debug": "2.2.0",
     "engine.io": "1.7.2",
-    "socket.io-parser": "2.3.1",
-    "socket.io-client": "1.5.1",
-    "socket.io-adapter": "0.4.0",
     "has-binary": "0.1.7",
-    "debug": "2.2.0"
+    "object-assign": "^4.1.0",
+    "socket.io-adapter": "0.4.0",
+    "socket.io-client": "1.5.1",
+    "socket.io-parser": "2.3.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -24,12 +24,13 @@
...
   },
   "dependencies": {
+    "debug": "2.2.0",
     "engine.io": "1.7.2",
-    "socket.io-parser": "2.3.1",
-    "socket.io-client": "1.5.1",
-    "socket.io-adapter": "0.4.0",
     "has-binary": "0.1.7",
-    "debug": "2.2.0"
+    "object-assign": "^4.1.0",
```
darrachequesne: Could you please pin the dependency here please?

perrin4869: Done! Missed it sorry!

jasperck:
Hi, just wanna ask why require `object-assign` usage rather than native `object.assign`,
is there any issue about shallow/deep clone? or just because the `polyfill`

perrin4869:
@jasperck node 0.10 and 0.12 are still being supported and they don't have native `Object.assign`. You can check Travis build logs, my first commit here used the native one and it passed for node 4 and 6 but failed for the older node... Hence why I introduced the polyfill. Maybe a note should be made to remove the dependency once node older than 4.0 stops being supported.

jasperck:
@perrin4869 Thanks for the feedback. Yep I saw the build and got it now. Adding a note for removing the dependency in the future sounds good.

perrin4869:
@jasperck maybe opening an issue as a remainder might do it, but considering the number of open issues right now, it would just get buried and forgotten...
@darrachequesne not to sound too impatient, but can we expect a patched release sometime soon? Thanks!

mck-:
Yes please! This is breaking for us too 😄

darrachequesne:
1.6.0 is out!

Regarding node 0.10 and 0.12, I think we'll deprecate those version in the next major bump (which will be eventually needed if we want to update ws).

mck-:
Excellent – thanks! I'll test it later today 👍

mck-:
Works great 👌

Issue #2762, opened 2016-11-24
socket.io.js served using http2 cannot load sourcemap due to possible routing error


SeinopSys:
### You want to:

* [x] report a *bug*
* [ ] request a *feature*

### Current behavior & steps to reproduce

Navigating to `https://ws.example.com:8667/socket.io/socket.io.js.map` opens the exact same file as `https://ws.example.com:8667/socket.io/socket.io.js`.

### Expected behavior

According to the `//# sourceMappingURL=socket.io.js.map` comment at the end of the file a source map should be loaded from the URL instead of the script itself.

### Setup
- OS: Debian Jessie 8.6
- browser: Chrome 54.0.2840.99
- socket.io version: 1.6.0

### Other information (e.g. stacktraces, related issues, suggestions how to fix)

Most likely an issue with a match check that returns when it finds the name of the js file and ignores the `.map` at the end. I'm adding my server setup code as well in case it matters:

```js
var Server = require('socket.io'),
    express = require('express'),
    app = express(),
    https = require('http2'),
    cors = require('cors');

// CORS
app.use(cors(function(req, callback){
    var corsOptions = { origin: false };
    if (/^https:\/\/example\.com/.test(req.header('Origin')))
        corsOptions.origin = true;
    callback(null, corsOptions);
}));

app.get('/', function (req, res) {
    res.sendStatus(403);
});

var lex = require('letsencrypt-express').create({
    server: 'https://acme-v01.api.letsencrypt.org/directory',
    email: 'user@example.com',
    agreeTos: true,
    approveDomains: [ 'ws.example.com'],
});
var server = https.createServer(lex.httpsOptions, lex.middleware(app));
server.listen(PORT);
var io = Server.listen(server);
```

darrachequesne:
Hi! The sourcemap is not currently served (as of 1.6.0), but will be in the next release: https://github.com/socketio/socket.io/pull/2482.

SeinopSys:
Will the JS file be served minified, or will it be kept the same way as it is now?

darrachequesne:
I guess we can serve the minified file, too. Let's do this!

PR #2764, opened 2016-11-25
Update index.html to correct socket.io.js CDN URI

pranayshirolkar:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
the existing reference to the socket.io.js in the script tag doesn't exist anymore. That had the demo chat application not go ahead and kept asking what's your nickname?

### New behaviour
the new script tag refers to the CDN location of the socket.io.js file and the error will not be there anymore.

the existing reference to the socket.io.js doesn't exist anymore so updated it with the recent CDN backed version

### DIFF
--- a/examples/chat/public/index.html
+++ b/examples/chat/public/index.html
@@ -22,7 +22,7 @@
   </ul>
 
   <script src="https://code.jquery.com/jquery-1.10.2.min.js"></script>
-  <script src="/socket.io/socket.io.js"></script>
+  <script src="https://cdn.socket.io/socket.io-1.4.5.js"></script>
   <script src="/main.js"></script>
 </body>
 </html>


darrachequesne:
Hi! The `socket.io.js` file should be served, what are the steps to reproduce please?

Note: the `socket.io.js.map` is indeed not currently served (as of `v1.6.0`), but will be in the next release (https://github.com/socketio/socket.io/issues/2762)

pranayshirolkar:
I realized my mistake shortly after. The demo on chat.socket.io <http://chat.socket.io/> doesn’t function and can’t go past the “What’s your nickname?” page. It has an CHUNK ERROR requesting the socket.io.js file and socket = io(); line in the main.js also throws an unknown identifier error.

Thanks!

PR #2766, opened 2016-11-27
Update client location and serve minified file

darrachequesne:
Following https://github.com/socketio/socket.io-client/pull/1025

Closes https://github.com/socketio/socket.io/issues/2762


julianlam:
As indicated in the referenced bug, this is a breaking change and should have triggered a major version bump.

Issue #2771, opened 2016-11-30
Socket.use error drops server


serhiisol:
### You want to:

* [x] report a *bug*

### Current behaviour
Socket.use drops server in case if we'll pass error into the next function

### Steps to reproduce (if the current behaviour is a bug)
Next snippet reproduces error easily:

```javascript
let io = require('socket.io')(3000);

io.on('connection', (socket) => {
  socket.use((packet, next) => {
    next(new Error('Failed error, which drops server'));
  });
  socket.on('message', (message) => {
    console.log(message);
  });
});

```
### Expected behaviour
Should send error packet back to the client and definitely shouldn't drop server 

### Setup
- OS: macOS Sierra 10.12.1
- browser: Chrome Version 54.0.2840.98 (64-bit)
- socket.io version: 1.7.1

### Other information (e.g. stacktraces, related issues, suggestions how to fix)

```
events.js:165
      throw err;
      ^

Error: Uncaught, unspecified "error" event. (undefined)
    at Socket.emit (events.js:163:17)
    at Socket.emit (/Users/serhiysolonko/Google Drive/Projects/Code/@decorators/playground/socket/node_modules/socket.io/lib/socket.js:141:10)
    at /Users/serhiysolonko/Google Drive/Projects/Code/@decorators/playground/socket/node_modules/socket.io/lib/socket.js:501:21
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)
```

### Possible place to debug
* https://nodejs.org/api/events.html#events_error_events
* https://github.com/socketio/socket.io/blob/master/lib/socket.js#L495

### Result
Socket cannot use event ```error``` as fallback =), if I'll use ```socket.error``` instead, system works as expected

PR #2772, opened 2016-11-30
fix(socket): Fixes socket.use error packet which drops nodejs due to …

serhiisol:
This PR fixes https://github.com/socketio/socket.io/issues/2771

* [x] a bug fix

### Current behaviour
Socket.use middleware drops nodejs process if error thrown in next function

### New behaviour
Socket.use middleware sends ```error``` packet correctly without dropping nodejs instance.

Socket cannot emit ```error``` event, because default behavior of EventEmitter will drop nodejs instance.
Details: https://nodejs.org/api/events.html#events_error_events

### DIFF
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -498,7 +498,7 @@ Socket.prototype.dispatch = function(event){
   this.run(event, function(err){
     process.nextTick(function(){
       if (err) {
-        return self.emit('error', err.data || err.message);
+        return self.error(err.data || err.message);
       }
       emit.apply(self, event);
     });
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -498,7 +498,6 @@ Socket.prototype.dispatch = function(event){
   this.run(event, function(err){
     process.nextTick(function(){
       if (err) {
-        self.emit('error', err.data || err.message);
         return self.error(err.data || err.message);
       }
       emit.apply(self, event);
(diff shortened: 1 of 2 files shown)


serhiisol:
This PR fixes the problem with missing error event on client (real client, not simulation in tests) but still you have to register at least one callback, due to EventEmitter core logic

darrachequesne:
Hi! Thanks for the PR. Actually, there seems to be an issue in the [tests](https://github.com/socketio/socket.io/blob/master/test/socket.io.js#L2271):

```js
it('should pass errors', function(done){
  var srv = http();
  var sio = io(srv);

  srv.listen(function(){
    var socket = client(srv, { multiplex: false });

    socket.emit('join', 'woot');

    // + socket.on('error', function(err){
    // +   expect(err).to.be('Authentication error');
    // +  done();
    // + });

    sio.on('connection', function(socket){
      socket.use(function(event, next){
        next(new Error('Authentication error'));
      });
      socket.use(function(event, next){
        done(new Error('nope'));
      });

      socket.on('join', function(){
        done(new Error('nope'));
      });
      // - socket.on('error', function(err){
      // -   expect(err).to.be('Authentication error');
      // -   done();
      // - });
    });
  });
});
```

According to the [docs](https://github.com/socketio/socket.io#socketusefnfunctionsocket):
> Errors passed to middleware callbacks are sent as special error packets to clients.

Your first commit should be sufficient I think. Could you please update the tests, by renaming the 2nd `socket`?

serhiisol:
@darrachequesne Hi, I noticed that after deep debugging, and I've updated PR with fixes

PR #2790, opened 2016-12-14
0.9: Remove process.EventEmitter usage for Node 7.x

johnwchadwick:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
Socket.IO 0.9.x does not work in Node.JS 7.0 due to the removal of `process.EventEmitter`. A myriad of existing software depends on Socket.IO 0.9.x, therefore it will either be necessary to update Socket.IO 0.9.x, or fork it. Some existing software still using 0.9.x includes the Yez extension for Google Chrome + corresponding Node.JS application.

### New behaviour
This patch changes usages of `process.EventEmitter` to `require('events')`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: node_js
 node_js:
-  - 0.6
+  - 0.12
+
+install:
+  - npm install -g npm
+  - npm install
 
 notifications:
   irc: "irc.freenode.org#socket.io"
--- a/examples/irc-output/irc.js
+++ b/examples/irc-output/irc.js
@@ -39,7 +39,7 @@ var Client = irc.Client = function(host, port) {
   this.user = null;
   this.real = null;
 }
-sys.inherits(Client, process.EventEmitter);
+sys.inherits(Client, require('events'));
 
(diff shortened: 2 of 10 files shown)


darrachequesne:
Hi! Would you know why the travis build fails?

johnwchadwick:
Not really sure. My guess would be that the ancient version of Node in this branch is not new enough to install the dependencies anymore. Should I bump it to something newer? 0.12, maybe?

darrachequesne:
Did you try with 0.10 or 0.12? It seems the tests are run with nodejs `0.6`..

johnwchadwick:
It's proving difficult to try to finagle things into a working state. Socket.IO depends on Expresso, a library that has been dead since around Socket.IO 0.9's time. I'm also seeing trouble with building Microtime, even if I update the package.json, which seems odd since it switched to `node-gyp` in 0.3.0.

thegiantbeast:
+1 for patching this issue.

darrachequesne:
Merged as 9ad1fd2771d641d17601ac7817c93dd63f3eec63, and released as 0.9.18. Thanks!

PR #2797, opened 2016-12-30
Use fs.readFileSync instead of require.resolve

a-lucas:

* [ ] a bug fix
* [ ] a new feature
* [x] other

### Current behaviour

Browserify doesn't supports require.resolve, and as a consequence, makes nexe fails the compilation : 

### New behaviour

This PR attempts to read the content of the socket.io-client file via fs.readFileSync and falls back to the original require.resolve if this file cannot be found. 

https://github.com/nexe/nexe/issues/289

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -5,6 +5,8 @@
 
 var http = require('http');
 var read = require('fs').readFileSync;
+var path = require('path');
+var exists = require('fs').existsSync;
 var engine = require('engine.io');
 var client = require('socket.io-client');
 var clientVersion = require('socket.io-client/package').version;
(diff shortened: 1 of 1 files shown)


darrachequesne:
Hi, thanks for the PR! How about using the following:

```js
function resolvePath(file){
  var filepath = path.resolve(__dirname, './../../', file);
  if (exists(filepath)) {
    return filepath;
  }
  return require.resolve(file);
}
```
Can you confirm the client file gets properly served that way? (you can also disable serving the client file with `serveClient: false`)

darrachequesne:
Else, [brfs](https://github.com/substack/brfs) may be helpful here, if you actually want to serve the file.

a-lucas:
@darrachequesne I never heard of brfs before, I also reckon your first suggestion is the best one, much DRYer, and also it doesn't requires an additional dependency.

Yes it works, [nexe](https://github.com/nexe/nexe) works great with socket.io with this change.

darrachequesne:
Thanks!

a-lucas:
Thanks too! Socket.io is great!

khuongduybui:
Has this been released? I'm using socket.io 2.0.4 and is still plagued with "require.resolve" calls.

Bolandish:
I'm also getting this error..

PR #2826, opened 2017-01-14
Reset properties instead of deleting them

Nibbler999:
Deleting properties is a v8 'optimization killer'

* [ ] a bug fix
* [ ] a new feature
* [ ] other


darrachequesne:
Thanks!

Out of curiosity, do you have any good resource documenting those 'optimization killers' (besides the one [here](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers))?

Issue #321, opened 2011-06-30
JSON flag reset


renajohn:
I'm using ZeroMQ and I've binded my ZeroMQ socket with my Socket.IO client in the following way:

``` javascript
// client is a socket.io client
client.__listener = client.json.send.bind(client);

zmqSocket.on("message", client.__listener);      
```

This use to work just fine with version 0.6, where I was calling `client.__listener = client.send.bind(client)`.

Here is the workaround I'm using:

``` javascript
client.__listener = function(msg){
   client.json.send(msg);
};  

zmqSocket.on("message", client.__listener);          
```

Any idea why the first snipped doesn't work?

Thanks,
Renault

PR #2838, opened 2017-01-26
Add the hasBinary flag

Andrews54757:

* [ ] a bug fix
* [X] a new feature
* [X] an update to the documentation
* [ ] other

### Current behaviour

When an object is emitted, the hasBin() function is used which iterates through the object checking for binary data. (https://github.com/socketio/has-binary)

### New behaviour

The hasBinary flag allows for you to specify whether the object you emit has binary data or not, which skips the inefficient hasBin() function.  You can now do

```
socket.hasBinary(true).emit("hello",someobjectwithbinary); // Contains binary data

socket.hasBinary(false).emit("helloagain",objectwithnobin); // No binary data

socket.emit("bye",anotherobject); // normal function
```

I found out that I could make socket.io two times faster by implementing this. Since in many circumstances a developer would know when he/she will send binary data (or not), the performance can be drastically improved by providing this information. This is because without it, the HasBinary function will run, which itinerates through everything in a object to check for binary data. This is very inefficient with large objects. This new flag allows for developers to fix this problem.


darrachequesne:
Hi Andrew,

Thanks for the pull request! From what I understand (and what that [comment](https://github.com/socketio/socket.io/issues/321#issuecomment-1500004) suggests), the `json` flag defined [here](https://github.com/socketio/socket.io/blob/1.7.2/lib/socket.js#L41) and [here](https://github.com/socketio/socket.io/blob/1.7.2/lib/namespace.js#L33) was initially meant to be used that way (the payload was `JSON.stringify`ed when the flag was present).

It does not seem to be used anymore, but it's a good use-case here, what do you think? How about something like:
```js
type: !this.flags.json && hasBin(args) ? parser.BINARY_EVENT : parser.EVENT
```
(in both places)

Andrews54757:
@darrachequesne Hi, I dont think that would work, as one of the main features of the hasBinary flag is so you can set it to true or false. Both options would override the hasBin function. If I had a large object containing a binary, the JSON flag wouldn't work as it does have a binary, so the HasBin function would then run, which I am trying to avoid. Also, the naming is less confusing if you use hasBinary, not JSON.


Thanks

darrachequesne:
@Andrews54757 Ok thanks, I understand. A few remarks though:

- would `withBinary` sound better, for chaining calls?
- could you apply the change to the `Namespace` class too? (`io.hasBinary(<bool>).emit`)
- would `(this.flags.hasBinary !== undefined ? this.flags.hasBinary : hasBin(args)) ? parser.BINARY_EVENT : parser.EVENT,` look clearer?
- could you update the documentation please?

Andrews54757:
@darrachequesne 

* `withBinary` has more characters than `hasBinary`, I personally enjoy `hasBinary` better.
* Done, I applied it to the namespace
* Done
* Done - Except that I am not good at writing documentations. Might need some review.

Swinkid:
Doc looks good to me - Anyone disagree? :)

docs/API.md
```diff
@@ -545,6 +555,17 @@ io.on('connection', function(socket){
 });
 ```
 
+#### Flag: 'hasBinary'
+
+Specifies whether there is binary data in the data to send. Increases preformance when specified. Can be `true` or `false`.
```
billouboq: preformance -> performance

docs/API.md
```diff
@@ -327,6 +329,14 @@ Sets a modifier for a subsequent event emission that the event data may be lost
 io.volatile.emit('an event', { some: 'data' }); // the clients may or may not receive it
 ```
 
+#### Flag: 'hasBinary'
+
+Specifies whether there is binary data in the data to send. Increases preformance when specified. Can be `true` or `false`.
```
billouboq: preformance -> performance

darrachequesne:
Thanks for the updates! I am still not convinced about the `hasBinary` name though, anyone else thinking it looks like it will return a boolean? Is it just me?

Swinkid:
@darrachequesne Two minds about it - think something like containsBinary would be better - but I then feel that is too long.

Andrews54757:
I personally like hasBinary better because it is short and concise.

Andrews54757:
@darrachequesne How about we expand this further. So it would be `method()` instead. where types are specified, such as `method("binary")` or `method("string")`. Then, this could be used in the future when there are other available methods. (or `parseMethod`?)

Andrews54757:
However, I feel that it would be performance issue to check strings. Any ideas?

darrachequesne:
Closed in favor of https://github.com/socketio/socket.io/pull/2829 and https://github.com/socketio/socket.io/pull/2923.

wagenaartje:
Shame this wasn't merged. This single function is what is slowing my server down to a whole other level as i'm sending large JSON's. The `socket.io-parser` that @darrachequesne implemented is vague, especially with no documentation (except for a few examples).

I simply replaced all the code inside `hasBinary` to `return false` and it made my server perform so much better.

Andrews54757:
@wagenaartje Yes, I agree with you. Thats why I stopped using socket.io and built my own from scratch.

darrachequesne:
Merged as https://github.com/socketio/socket.io/pull/3185, sorry for the delay.

PR #2846, opened 2017-01-31
Small addition to the Express Readme Part

H3rby7:

* [ ] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
I used the express setup with my latest project and stumbled into an error, that was hard to fix, because it was so small and crucial.  
I accidently called `app.listen(PORT)` instead of `server.listen(PORT)`  
This basically leads to working Express functionality and routes, but the `/socket.io` route will not be present and thus the GET request for any client willing to connect gets a 404.  

### New behaviour
I hope this small comment may save another one the four hours I spent on tracing my mistake.  


darrachequesne:
Thanks for the contribution!

PR #2853, opened 2017-02-08
fix mismatch between code example and docs

timruffles:
This is documentation for the server code, and the commented code demonstrates the client's response.

Update the wording to match.

* [ ] a bug fix
* [ ] a new feature
* [ ] other


darrachequesne:
Good catch, thanks!

PR #2873, opened 2017-02-23
Make Socket, Namespace inherit EventEmitter properly

i8-pi:

Setting an object's prototype is discouraged, and `__proto__` is not required to exist outside of browsers.

Use the standard way of making a class inherit from another

### DIFF
--- a/lib/namespace.js
+++ b/lib/namespace.js
@@ -50,6 +50,7 @@ var emit = Emitter.prototype.emit;
  */
 
 function Namespace(server, name){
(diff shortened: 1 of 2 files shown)


shirtleton:
This would be better solved with ES6 classes as proposed earlier.
#2514

darrachequesne:
Closed by https://github.com/socketio/socket.io/commit/3d760b71d7ee480dfab4b9f336a5d90d970c6b3f.

PR #2874, opened 2017-02-23
Explicitly document that Server extends EventEmitter

i8-pi:

Explicitly say that Server extends EventEmitter, and list a few methods from EventEmitter that are potentially useful.


darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/240b154960d89e65ccddddf4319f07e9e63d69b4. Thanks a lot!

PR #2889, opened 2017-03-07
made chat example have a responsive layout

jchannon:

* [] a bug fix
* [ ] a new feature
* [x] other

### Current behaviour
Chat example is not responsive

### New behaviour
Chat example is responsive

Introduces a minimal css reference to skeleton css fwk around 1kb.

### DIFF
--- a/examples/chat/public/index.html
+++ b/examples/chat/public/index.html
@@ -1,28 +1,39 @@
 <!doctype html>
 <html lang="en">
+
 <head>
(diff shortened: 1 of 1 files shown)


darrachequesne:
I'm not sure that it really adds value here, thanks.

jchannon:
Added value for me so I could test scoket.io on a phone/tablet but each to their own 😄

PR #2895, opened 2017-03-10
Add note in docs for `origins(fn)` about `error` needing to be a string.

sgress454:

* [ ] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour

The docs for `origins(fn)` currently don't mention that `error` has to be supplied as a string.  This really tripped me up as I was supplying an Error instance as I usually do for callbacks, which crashes w/ certain versions of Node.

### New behaviour

Added line to documentation indicating that `error` should be a string.

An alternative to this would be to patch `engine.io` to allow an Error instance in its [abortConnection method](https://github.com/socketio/engine.io/blob/bd1e81ec3d11f18c5fc09c55745d5ff66d037cc1/lib/server.js#L497), using its `message` property as the error message.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -155,7 +155,7 @@ Sets the allowed origins `value`. Defaults to any origins being allowed. If no a
   - `fn` _(Function)_
   - **Returns** `Server`
 
-Provides a function taking two arguments `origin:String` and `callback(error, success)`, where `success` is a boolean value indicating whether origin is allowed or not.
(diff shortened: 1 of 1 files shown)


YvonneMcLeod:
Pull/2895

darrachequesne:
Thanks :+1: 

That behaviour is indeed weird, I'll add it to the list of breaking changes for version 4.0.0 of engine.io.

PR #2906, opened 2017-03-31
[docs] Update emit cheatsheet

FarazPatankar:

* [ ] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour

- The information currently present in the cheat sheet does not show users how they can emit to a room that is within a namespace.

### New behaviour

- Adds an example of how someone would emit to a room within a namespace.


darrachequesne:
Thanks!

PR #2907, opened 2017-04-01
add cache-control header to client source

maxlego:

* [ ] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
No cache-control header is sent when requesting clien script

### New behaviour
Server will send cache-control header with value "public, max-age=0"


darrachequesne:
Finally! Thanks :clap:

PR #2910, opened 2017-04-17
fix(Node v6.9.1) process.EventEmitter was deprecated

lcamilo15:
Make version v0.9 compatible with Node ^v6.9.1 by using require('events') instead of process.EventEmitter

* [x] a bug fix
* [x] a new feature
* [ ] other


darrachequesne:
Released as 0.9.18 :+1:

PR #2913, opened 2017-04-18
Fixed "Getting Started" not being bold

epicTCK:
It had showed four pound signs, now it is bold. I fixed it by deleting and re-adding the space in between the pounds and the text. Not sure why that was the issue, but I fixed it nonetheless.

* [ ] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
Showed four pound signs

### New behaviour
It's bold now

None


darrachequesne:
@epicTCK thanks!

PR #2933, opened 2017-05-08
Activating Open Collective

xdamman:
Please merge this pull request to activate your Open Collective https://opencollective.com/socketio

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1,6 +1,7 @@
 
 # socket.io
 
+[![Backers on Open Collective](https://opencollective.com/socketio/backers/badge.svg)](#backers) [![Sponsors on Open Collective](https://opencollective.com/socketio/sponsors/badge.svg)](#sponsors)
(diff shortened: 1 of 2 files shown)


darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/0d07c47f815e2174dcafac29441123682677313c.

I'm not really fond of the `postinstall` script, so I excluded it from the merge. Is it something mandatory?

xdamman:
It's not but it's a nice touch to ask for donations. Have you tried it? You can also pass it the `--plain` option to make it text only (and more subtle)

darrachequesne:
@xdamman I am still not convinced that adding another dependency (something like 49 Mo if I'm not mistaken) is really necessary here. Besides, what will happen if webpack and babel and mocha and gulp... all add that postinstall script?

xdamman:
I see 22MB on my machine. But that's highly variable depending on what other modules you already have installed in your project.

If other projects also have it, npm is smart enough to only install it once.

Issue #2935, opened 2017-05-10
Server lost the message which is sent in client reconnect event handler


mingc00:
*Note*: for support questions, please use one of these channels: [stackoverflow](http://stackoverflow.com/questions/tagged/socket.io) or [slack](https://socketio.slack.com)

### You want to:

* [x] report a *bug*
* [ ] request a *feature*

### Current behaviour
We register asynchronous middleware for authentication. After upgrading socket.io from 1.7.4 to 2.0.1, server cannot get the message if client sent it in `reconnect` event handler.

### Steps to reproduce (if the current behaviour is a bug)
Please reference the code: https://github.com/mingc00/socket.io-fiddle

1. Register a asynchronous middleware
2. Startup server
3. Turn off server after client is connected
4. Restart the server
5. Server cannot get the message which is sent in client `reconnect` event handler.

**Note**: the best way to get a quick answer is to provide a failing test case, by forking the following [fiddle](https://github.com/darrachequesne/socket.io-fiddle) for example.

### Expected behaviour
In previous version, server should get the message which is sent in client `reconnect` event handler.

### Setup
- OS: Ubuntu 16.04
- browser: Chrome 58
- socket.io version: 2.0.1

adrai:
same issue even if directly emitting a message in the `connect` event handler

darrachequesne:
@mingc00 thanks for the fiddle, I could indeed reproduce.

It seems to be related to https://github.com/socketio/socket.io/pull/2833, the CONNECT packet is sent too soon.

adrai:
same issue as: https://github.com/socketio/socket.io/issues/2866

mxq19901028:
How to ignore the LTS certificate problem?

PR #2948, opened 2017-05-20
fix timing issues with middleware

darrachequesne:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour

Using a middleware could previously lead to a connecting client
receiving a connect event from the server before the server triggers
its own connect event.

Fixes https://github.com/socketio/socket.io/issues/2935.

### DIFF
--- a/lib/namespace.js
+++ b/lib/namespace.js
@@ -99,6 +99,8 @@ Namespace.prototype.initAdapter = function(){
  */
 
 Namespace.prototype.use = function(fn){
+  debug('removing initial packet');
+  delete this.server.eio.initialPacket;
   this.fns.push(fn);
   return this;
 };
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -297,9 +297,10 @@ Socket.prototype.onconnect = function(){
   debug('socket connected - writing packet');
   this.nsp.connected[this.id] = this;
   this.join(this.id);
-  // the CONNECT packet for the default namespace
-  // has already been sent as an `initialPacket`
-  if (this.nsp.name !== '/') {
+  var skip = this.nsp.name === '/' && this.nsp.fns.length === 0;
+  if (skip) {
+    debug('packet already sent in initial handshake');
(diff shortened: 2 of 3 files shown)


AVVS:
this created another problem - if `.attach/listen` was not called before `.use` it crashes node process due to `delete this.server.eio.initialPacket;` as `this.server.eio` is undefined before that call

dolphinlin:
the commit let SWIFT SDK can't run!!!!!!!!!!!!!!
SWIFT connect to namespace that use a middleware, there is fail!!!

i delete the code, it works!!!!

```js
// delete this.server.eio.initialPacket;
```

### socket.io client
```swift
func testLocalhost() {
        let expectation : XCTestExpectation = self.expectation(description: "Wait for socket call back")
        
        let accessToken = "qqq"
        
        let socket = SocketIOClient(socketURL: URL(string:"http://localhost:3000")!, config: [
            .log(true),
            .nsp("/location"),
            .forceWebsockets(true),
            .connectParams(["x-access-token": accessToken]),
            .secure(false)
            ])
        
        socket.on("connect") { (data:[Any], emitter:SocketAckEmitter) in
            XCGLogger.info("We got connect.")
            expectation.fulfill()
        }
        socket.connect()
        
        self.waitForExpectations(timeout: waitForTimeout + 5) { (error:Error?) in
            if error != nil {
                XCGLogger.error("Waiting time out")
            }
        }

    }
```

### socket.io server
```js
var app = require('express')();
var server = require('http').createServer(app);
var io = require('socket.io')(server);
var locationIO = io.of('/location')

const authCheck = (client, next) => {
	console.log('auth check')
    console.log('Access Token:' + client.handshake.query['x-access-token'])
    next()
}

io.on('connection', () => {
	console.log('main io connection!')
})

locationIO.use(authCheck)

locationIO.on('connection', function(){ 
	console.log('client connect success!')
});


server.listen(3000);
```

PR #2960, opened 2017-06-03
Update webpack example

darrachequesne:

* [ ] a bug fix
* [ ] a new feature
* [ ] other


lib/index.js
```diff
@@ -8,8 +8,7 @@ var read = require('fs').readFileSync;
 var path = require('path');
 var exists = require('fs').existsSync;
 var engine = require('engine.io');
-var client = require('socket.io-client');
-var clientVersion = require('socket.io-client/package').version;
+var clientVersion = require('socket.io-client/package.json').version;
```
darrachequesne: Note: This change is not mandatory but it seems to be an issue sometimes (https://github.com/socketio/socket.io-client/issues/933#issuecomment-299776125).

PR #2987, opened 2017-06-23
Update main.js

VincentGau:
Fix the cleanInput() function to prevent input from having injected markup.

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
the function returns exactly the same input.

### New behaviour
prvents injected markup


j-f1:
I think you want `.text(input).html()`.

PR #3014, opened 2017-07-24
getting rid of github warning

Stanley288:
getting rid of github warning of using path.join


darrachequesne:
@Stanley288 thanks!

PR #3032, opened 2017-08-17
Adding Socket ID to disconnecting and disconnect event callbacks

nbcnc:
Needed a way to determine which socket disconnected.  Updated the emit() callbacks for "disconnecting" and "disconnected" to return the unique ID of the socket connection.

* [ ] a bug fix
* [x] a new feature
* [ ]  other

### Current behaviour

    client.on('disconnect',()=>{
       // returns no useful information
    });

### New behaviour

    client.on('disconnect',(reason,id)=>{
        console.log(`The user with socket ID ${id} just disconnected`);
    });

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -96,7 +96,9 @@ var server = require('http').createServer();
 var io = require('socket.io')(server);
 io.on('connection', function(client){
   client.on('event', function(data){});
-  client.on('disconnect', function(){});
+  client.on('disconnect', function(id){
+    console.log("Socket with ID ",id," Disconnected");
+  });
 });
 server.listen(3000);
 ```
--- a/Readme.md
+++ b/Readme.md
@@ -96,7 +96,7 @@ var server = require('http').createServer();
 var io = require('socket.io')(server);
 io.on('connection', function(client){
   client.on('event', function(data){});
-  client.on('disconnect', function(id){
+  client.on('disconnect', function(reason, id){
     console.log("Socket with ID ",id," Disconnected");
   });
 });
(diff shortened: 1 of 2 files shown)


darrachequesne:
You should already be able to access the socket id:
```js
client.on('disconnect',(reason,id)=>{
    console.log(`The user with socket ID ${client.id} just disconnected`);
});
```

Am I missing something?

nbcnc:
As far as I can tell, the current "disconnect" event callback doesn't pass any parameters except "reason" -- which is just a string.  My proposal is to add "id" to the callback so that we know which client.id was disconnected.  The client.id is currently stored as this.id in the Socket class.  I'm merely proposing to pass that back as a convenience.

nbcnc:
just following up -- will this change get pulled into the master repo?

idibidiart:
io.socket.on('connection', function (socket) { // socket.on('disconnect',...)  happens here }) so you already have access to socket.id

shirtleton:
We're using something like @darrachequesne's example in production.  "Convenience" is relative here since the `id` is already available on the instance of `client` in your example.

dasilvacontin:
@nbcnc Following that logic, wouldn't we need to add the `id` to all the `.on` callbacks? Otherwise it would be inconsistent.

And then, wouldn't we need to also pass other properties that `client`/`socket` has? Why `id` only?

I think we should strive for consistency. This feels more like something that was useful in the particular case you mention. Maybe in a different case it would have been useful to get the namespaces or rooms the client is in. That is not a reason to be adding the parameter to the callback. Did I make sense?

PR #3078, opened 2017-10-09
Use npm scripts

thiamsantos:

* [ ] a bug fix
* [ ] a new feature
* [x] other

### Current behaviour
Currently it uses gulp and make.

### New behaviour
Use just plain and simple npm scripts.


darrachequesne:
@thiamsantos hi! That's indeed a good idea, I think we can even remove the babel transpilation for now, since the transpiled files are not used / published.

thiamsantos:
@darrachequesne Done. Babel transpilation removed.

darrachequesne:
@thiamsantos great, thanks a lot!

Issue #3139, opened 2017-12-18
CreateListFromArrayLike called on non-object


kaansoral:
### You want to:

* [x] report a *bug*
* [ ] request a *feature*

### Current behaviour

Server dies instantly, cause unknown

### Steps to reproduce (if the current behaviour is a bug)

This happened passively, I found this in my error log

```
/home/ubuntu/node_modules/socket.io/lib/socket.js:503
      emit.apply(self, event);
           ^

TypeError: CreateListFromArrayLike called on non-object
    at /home/ubuntu/node_modules/socket.io/lib/socket.js:503:12
    at _combinedTickCallback (internal/process/next_tick.js:73:7)
    at process._tickCallback (internal/process/next_tick.js:104:9)
```

### Expected behaviour

Handle the issue, provide more information

### Setup
- OS: Ubuntu
- browser: Node 7.7.1
- socket.io version: 1.7.3

hengwoon:
I have been seeing a bunch of these errors as well

Chopinsky:
`apply` expects `event` to be a type of array, if `event` is of any other data type, `apply` will throw the TypeError. My guess is that event is somewhat null or undefined in this case, probably because decoding data failed and the `packet.data` is in ill form. Anyway, we should probably check if event is an array before calling `apply` with it as 2nd argument.

I will submit a PR to put a safeguard in place.

jamesbachini:
I had this issue on a fairly high load server. Was getting a lot of uncaught exceptions. The way I fixed it was to add this line to /node_modules/socket.io/lib/socket.js

if (typeof event !== 'object') return false;

Above line 513,       emit.apply(self, event);

This seems to have fixed the issue for now but not sure if it's a good permanent solution.

JohnRSim:
I'm also experiencing the same issue..

zalemmm:
Same issue here,  after a migration of my projets to a new workstation,  and trying to launch my gulp tasks  over a fresh node.js & npm packages installation. I tried other solutions, downgrading node.js to my last machine version, erasing & reinstall of node modules, but only  jamesbachini  solution works for me

darrachequesne:
Fixed by https://github.com/socketio/socket.io-parser/commit/f0a7df10599d9f6ff392bdfb8352489802220fd3, included in `socket.io-parser@3.1.3` & `socket.io@2.0.4`.

PR #3140, opened 2017-12-19
Adding rail guard to dispatch's event param

Chopinsky:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
When event packet received by `onevent` doesn't contain a valid array as the value of the `data` property, the subsequent call to dispatch will cause a `TypeError`, because the `apply` call on `emit` would expect the 2nd parameter to be an array type. 

### New behaviour
Now if event packet's `data` property is not already an array, we will convert it to an array before passing it to the `apply` call, to avoid the `TypeError` exception on rare occurrence of the problem as raised in issue #3139.

There seem to be quite a few other places that would use a similar array type check in `socket.js`, but would rather hold off on updating all these places until further reporting of the `TypeError` exceptions.

### DIFF
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -503,8 +503,11 @@ Socket.prototype.compress = function(compress){
  */
 
 Socket.prototype.dispatch = function(event){
-  debug('dispatching an event %j', event);
+  event = Array.isArray(event) ? event : [event];
   var self = this;
+
+  debug('dispatching an event %j', event);
+
   function dispatchSocket(err) {
     process.nextTick(function(){
       if (err) {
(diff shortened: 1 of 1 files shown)


grajagandev:
Would you be interested in software that automatically finds TypeErrors like this? I am building [Fuzz Stati0n](https://fuzzstati0n.com?id=3140) to do that (free for OSS) - please take a look and consider signing up for our newsletter to keep in touch.

darrachequesne:
Closed by https://github.com/socketio/socket.io-parser/pull/83

PR #3154, opened 2018-01-04
fix cannot catch error where error in a listener

czonechan:

* [x] a bug fix

### Current behaviour
client.on('error', (err) => {
    console.log(err);
});

client.on('entter', () => {
    try {
       throw new Error('test');
   } catch (err) {
      client.emit('error', err);
  }    
});

i must try catch error in a listener,otherwise the error listener does not work.
### New behaviour
client.on('error', (err) => {
    console.log(err);
});
client.on('entter', () => {
    throw new Error('test');
});

### DIFF
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -510,7 +510,11 @@ Socket.prototype.dispatch = function(event){
       if (err) {
         return self.error(err.data || err.message);
       }
-      emit.apply(self, event);
+      try {
+          emit.apply(self, event);
+      } catch(e) {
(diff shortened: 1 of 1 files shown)


shirtleton:
I don't think this is a bug. The current behavior allows developers to do their own error handling.
Personally, I'm not a fan of all of the errors going through one event listener.

darrachequesne:
I think @shirtleton is right, the developers have to do their own error handling.

PR #3195, opened 2018-03-08
 feat: add support for dynamic namespaces (WIP)

darrachequesne:

* [ ] a bug fix
* [x] a new feature
* [ ] other

### New behaviour

The following test should describe the expected behaviour:
```js
const socket = client(srv, '/dynamic-101');
io.of(/^\/dynamic-\d+$/).on('connect', (socket) => {
  // socket.nsp.name = '/dynamic-101'
});

socket.on('error', function(err) {
  expect().fail();
});
socket.on('connect', () => {
  // ...
});
```


kinesivan:
Is there a way to pass data into next when using a function to create dynamic namespaces?

Issue #3217, opened 2018-04-04
Broadcast to local instance only when using the redis-adapter


sleimanx2:
it's possible to add the local flag when emitting to a room

```js
 // sending to all clients on this node (when using multiple nodes)
 io.to(channel).local
     .emit(message.event, channel, message.data)
```

however, when broadcasting from a socket instance it's not possible to add a local flag is there any way to achieve the same behavior?

```js
// this wont work
socket.broadcast.to(channel).local
      .emit(message.event, channel, message.data);
```
thank you

darrachequesne:
Closed by https://github.com/socketio/socket.io/pull/3219

PR #3219, opened 2018-04-06
add local flag to the socket object

sleimanx2:

* [ ] a bug fix
* [ ] a new feature
* [x] other

### Current behaviour

Currently, the only way to set the local flag when in socket context is as follow

```js
socket.flags['local'] = true;
```

### New behaviour

Now we can add the local flag the same way we do it on the Namespace object

```js
socket.local...
```

https://github.com/socketio/socket.io/issues/3217

### DIFF
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -39,7 +39,8 @@ exports.events = [
 var flags = [
   'json',
   'volatile',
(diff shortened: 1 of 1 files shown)


darrachequesne:
Thanks!

PR #3234, opened 2018-04-28
[chore] add JSON extension to require

oayres:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
Requires package JSON of socket.io/client without the JSON extension defined. This causes those of us who bundle this dependency to bump into a failure, as it cannot resolve the file. This is already sorted in the later versions of socket.io, but we're unfortunately tied to 1.7.4 for now.

### New behaviour
Adds JSON extension in the require so it can resolve the dependency when bundling.

We're using Fusebox as our module bundler. If this can make it into a 1.7.5 patch, that'd be ace.


darrachequesne:
Thanks!

PR #3235, opened 2018-05-03
Add the "loopback" flag.

ripfester:

* [ ] a bug fix
* [x] a new feature
* [ ] other

### Current behaviour

It is not possible to send messages to the local (server) endpoint. In scenarios where rooms/broadcasts are used, this requires the server to maintain a separate room->socket mapping so that administrative actions can be performed (e.g., close all sockets in room 'foo').

### New behaviour

With this change (and related PR https://github.com/socketio/socket.io-adapter/pull/56), it is possible to emit/broadcast messages that are handled by the local (server) endpoint by inserting `.loopback` into the emit syntax. On connect, the server can add handlers for messages that originate server-side, and later broadcast those messages in order to effect administrative actions (e.g., close all sockets in room 'foo') without maintaining independent mappings.

https://github.com/socketio/socket.io-adapter/pull/56 must be committed before this change, as `socket.io-adapter` is used in tests and must honour the `loopback` flag.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -163,6 +163,11 @@ Client.prototype.packet = function(packet, opts){
   opts = opts || {};
   var self = this;
 
+  if ('open' != this.conn.readyState) {
+    debug('ignoring packet write %j', packet);
+    return;
+  }
+
   // this writes to the actual connection
   function writeToEngine(encodedPackets) {
     if (opts.volatile && !self.conn.transport.writable) return;
@@ -171,15 +176,23 @@ Client.prototype.packet = function(packet, opts){
     }
   }
 
-  if ('open' == this.conn.readyState) {
+  if (opts.loopback) {
+    debug('writing loopback packet %j', packet);
+    // handle loopback packets as though they came in off the wire.
+    if (opts.preEncoded) { // a broadcast pre-encodes a packet
+      for (var i = 0; i < packet.length; i++) {
+        this.decoder.add(packet[i]);
+      }
+    } else { // no broadcasting, no need to decode
+      this.ondecoded(packet);
+    }
+  } else {
     debug('writing packet %j', packet);
     if (!opts.preEncoded) { // not broadcasting, need to encode
       this.encoder.encode(packet, writeToEngine); // encode, then write results to engine
     } else { // a broadcast pre-encodes a packet
       writeToEngine(packet);
     }
-  } else {
-    debug('ignoring packet write %j', packet);
   }
 };
 
--- a/lib/namespace.js
+++ b/lib/namespace.js
@@ -30,6 +30,7 @@ exports.events = [
  */
 
 exports.flags = [
+  'loopback',
   'json',
   'volatile',
   'local'
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -37,6 +37,7 @@ exports.events = [
  */
 
 var flags = [
+  'loopback',
   'json',
   'volatile',
   'broadcast',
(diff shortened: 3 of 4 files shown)


chrisleck:
Would it be possible to pull this in?  Would be useful.

darrachequesne:
> this requires the server to maintain a separate room->socket mapping so that administrative actions can be performed (e.g., close all sockets in room 'foo').

Couldn't you just do the following?

```js
const room = io.of('/').adapter.rooms['foo'];
const sockets = room.sockets;

for (var id in sockets) {
  if (sockets.hasOwnProperty(id)) {
    const socket = io.of('/').connected[id];
    if (socket) {
      socket.disconnect();
    }
  }
}
```

Well, you're accessing the adapter internals, but that should work, shouldn't it?

chrisleck:
That won't work when you're using something like `socket.io-redis`, will it?

ripfester:
It's been a couple of years, but if I recall correctly @chrisleck is correct. This functionality was specifically desired to allow these types of server-side actions to be generated on a single server and distributed/handled across multiple servers when using the redis adapter.

darrachequesne:
Would [customHook/customRequest](https://github.com/socketio/socket.io-redis#redisadaptercustomrequestdataobject-fnfunction) from the Redis adapter suit your needs? (there's not much documentation about this feature, I'll try to fix this)

chrisleck:
Hi @darrachequesne, I took a look at the customHook/customRequest function, but it looks like that doesn't allow for sending to individual rooms?  Or am I reading that incorrectly?

chrisleck:
Hi @darrachequesne, I'm not trying to be annoying, but would like to make some forward progress here if possible. Any feedback on my comment above regarding sending to rooms through the custom* functions you pointed out above?

chrisleck:
Any updates @darrachequesne ?

chrisleck:
Hello folks - any updates on taking this PR?

darrachequesne:
@chris-codaio sorry for the delay.

If I correctly understand the suggested change, there is no way to make a distinction between packets sent by the client-side and those sent by the server-side, which seems problematic. Or am I missing something?

[This commit](https://github.com/socketio/socket.io/commit/b25495c069031674da08e19aed68922c7c7a0e28) (which will be released in the next version of Socket.IO) adds the following methods:

```js
// return all Socket instances of the "admin" namespace in the "room1" room
const sockets = await io.of("/admin").in("room1").fetchSockets();

// make all Socket instances of the "admin" namespace in the "room1" room join the "room2" room
io.of("/admin").in("room1").socketsJoin("room2");

// make all Socket instances of the "admin" namespace in the "room1" room leave the "room2" room
io.of("/admin").in("room1").socketsLeave("room2");

// make all Socket instances of the "admin" namespace in the "room1" room disconnect
io.of("/admin").in("room1").disconnectSockets();
```

Would it suit your use case?

If it does not, I think we could indeed add the `loopback` flag, but at the namespace level instead:

```js
// main namespace
io.loopback.emit("hello");
io.on("hello", () => { /* ... */ });

// custom namespace
io.of("/admin").loopback.emit(/* ... */);
io.of("/admin").on(/* ...*/);

// maybe with acknowledgements too, in a multi server setup
io.loopback.emit("hello", 1, "2", (err, responses) => {
  // ...
});
io.on("hello", (arg1, arg2, cb) => {
  cb(42);
});
```

What is your opinion?

darrachequesne:
For future readers: this was implemented with the `serverSideEmit` feature.

Documentation: https://socket.io/docs/v4/server-api/#namespace-serverSideEmit-eventName-%E2%80%A6args-ack

PR #3240, opened 2018-05-07
Improve the chat example with more ES6 features

sergiolenza:
- Add default initializers
- Use let and const instead var

* [ ] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
The chat example can be updated with ES6 features like default params, constants (avoiding the use of `var`), and template literals.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

### New behaviour
The chat example with these three features.

### DIFF
--- a/examples/chat/public/main.js
+++ b/examples/chat/public/main.js
@@ -1,36 +1,36 @@
 $(function() {
-  var FADE_TIME = 150; // ms
-  var TYPING_TIMER_LENGTH = 400; // ms
-  var COLORS = [
+  const FADE_TIME = 150; // ms
+  const TYPING_TIMER_LENGTH = 400; // ms
+  const COLORS = [
     '#e21400', '#91580f', '#f8a700', '#f78b00',
     '#58dc00', '#287b00', '#a8f07a', '#4ae8c4',
     '#3b88eb', '#3824aa', '#a700ff', '#d300e7'
   ];
 
   // Initialize variables
-  var $window = $(window);
-  var $usernameInput = $('.usernameInput'); // Input for username
-  var $messages = $('.messages'); // Messages area
-  var $inputMessage = $('.inputMessage'); // Input message input box
+  const $window = $(window);
+  const $usernameInput = $('.usernameInput'); // Input for username
+  const $messages = $('.messages');           // Messages area
+  const $inputMessage = $('.inputMessage');   // Input message input box
 
-  var $loginPage = $('.login.page'); // The login page
-  var $chatPage = $('.chat.page'); // The chatroom page
+  const $loginPage = $('.login.page');        // The login page
+  const $chatPage = $('.chat.page');          // The chatroom page
+
+  const socket = io();
 
   // Prompt for setting a username
-  var username;
-  var connected = false;
-  var typing = false;
-  var lastTypingTime;
-  var $currentInput = $usernameInput.focus();
-
-  var socket = io();
+  let username;
+  let connected = false;
+  let typing = false;
+  let lastTypingTime;
+  let $currentInput = $usernameInput.focus();
 
   const addParticipantsMessage = (data) => {
-    var message = '';
+    let message = '';
     if (data.numUsers === 1) {
-      message += "there's 1 participant";
+      message += `there's 1 participant`;
     } else {
-      message += "there are " + data.numUsers + " participants";
+      message += `there are ${data.numUsers} participants`;
     }
     log(message);
   }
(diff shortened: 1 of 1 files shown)


examples/chat/public/main.js
```diff
@@ -118,19 +114,8 @@ $(function() {
...
-    if (!options) {
-      options = {};
-    }
-    if (typeof options.fade === 'undefined') {
-      options.fade = true;
-    }
-    if (typeof options.prepend === 'undefined') {
-      options.prepend = false;
-    }
+  const addMessageElement = (el, options = { fade: true, prepend: false }) => {
```
bksahu: 'X is typing' message not fading out.
I tried converting all the arrow function to normal function and it worked. But it's not working in arrow function.

examples/chat/public/main.js
```diff
@@ -53,45 +53,41 @@ $(function() {
...
-    const log = (message, options) => {
-    var $el = $('<li>').addClass('log').text(message);
+  const log = (message, options) => {
+    const $el = $('<li>').addClass('log').text(message);
     addMessageElement($el, options);
   }
 
   // Adds the visual chat message to the message list
-  const addChatMessage = (data, options) => {
+  const addChatMessage = (data, options = {}) => {
```
senyaak: This won't work if option is null.

examples/chat/public/main.js
senyaak: Also wont work if null will be passed

darrachequesne:
@sergiolenza could you please review the comments made to the PR? Thanks a lot!

sergiolenza:
> @sergiolenza could you please review the comments made to the PR? Thanks a lot!

Yes of course @darrachequesne.
I'll review it this weekend.
Thanks! 🙏

examples/chat/public/main.js
sergiolenza: 👌

examples/chat/public/main.js
sergiolenza: 👌

examples/chat/public/main.js
```diff
@@ -1,36 +1,36 @@
...
+  let typing = false;
+  let lastTypingTime;
+  let $currentInput = $usernameInput.focus();
 
   const addParticipantsMessage = (data) => {
-    var message = '';
+    let message = '';
     if (data.numUsers === 1) {
-      message += "there's 1 participant";
+      message += `there's 1 participant`;
```
dasilvacontin, line 31: The string here should be single quotes, since there's no variable being insert (following the style of the rest of the page).

```suggestion
      message += 'there's 1 participant';
```

examples/chat/public/main.js
```diff
@@ -53,45 +53,41 @@ $(function() {
...
   // Log a message
-    const log = (message, options) => {
-    var $el = $('<li>').addClass('log').text(message);
+  const log = (message, options) => {
+    const $el = $('<li>').addClass('log').text(message);
     addMessageElement($el, options);
   }
 
   // Adds the visual chat message to the message list
   const addChatMessage = (data, options) => {
```
dasilvacontin, line 75: There's a deleted line below that has no matching new line: `options = options || {};`.

Did you mean to add a default parameter for `options` here?
```suggestion
  const addChatMessage = (data, options = {}) => {
```

examples/chat/public/main.js
```diff
@@ -119,8 +115,7 @@ $(function() {
   // options.prepend - If the element should prepend
   //   all other messages (default = false)
   const addMessageElement = (el, options) => {
-    var $el = $(el);
-
+    const $el = $(el);
     // Setup default options
     if (!options) {
       options = {};
```
dasilvacontin, line 121: I'd suggest moving this to a default function parameter value.

dasilvacontin: I don't feel to strongly about this change though, because there's additional "default options logic" that can't be moved into the parameter description, since you may define `fade` and not `prepend`, and you still want `prepend` to default to the value below (following the logic in this code).

examples/chat/public/main.js
```diff
@@ -229,7 +225,7 @@ $(function() {
   socket.on('login', (data) => {
     connected = true;
     // Display the welcome message
-    var message = "Welcome to Socket.IO Chat – ";
+    const message = 'Welcome to Socket.IO Chat – ';
```
dasilvacontin, line 228: (exhibit A of use of single quotes on string with non-parameter)

PR #3268, opened 2018-06-06
Doc improvement

enricop89:

* [] a bug fix
* [ x ] a new feature
* [ x ] an update to the documentation
* [ ] other

Hello everyone, I updated the disconnect reasons in the documentation. I extracted them from the source code of the library. Please let me know if everything is correct.

I would like to propose a more precise format for the disconnection reasons, something like this:

```json
"code": xxx,
"reason": "The reason of the error",
"message": "The message of the error"
```

With this format, it is easier to parse reasons and build logic upon the disconnection event.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -853,13 +853,13 @@ io.on('connection', (socket) => {
 
 Available options:
 
-| Reason | Motivation |
-| ------- | ---------------|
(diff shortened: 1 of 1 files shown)


darrachequesne:
Merged there: https://github.com/socketio/socket.io/pull/3268

Thanks a lot :+1:

PR #3318, opened 2018-08-27
1.7.x

i-Clyde:

* [ ] a bug fix
* [x] a new feature
* [ ] other

### New behaviour
It would be also great to do something like this:
`io.sockets.connected[{socketA, socketB, socketC, socketD}].join(roomtoken)`
`io.sockets.connected[{socketA, socketB, socketC, socketD}].leave(roomtoken)`
That will make lobby based applications easier to manage ^^


darrachequesne:
Hi! Could you please describe the feature you want to add? (the commits seem unrelated)

PR #3321, opened 2018-09-02
Add badge to display install size

styfle:

* [ ] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour

No badge

### New behaviour

Badge that displays the npm install size

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -7,6 +7,7 @@
 [![devDependency Status](https://david-dm.org/socketio/socket.io/dev-status.svg)](https://david-dm.org/socketio/socket.io#info=devDependencies)
 [![NPM version](https://badge.fury.io/js/socket.io.svg)](https://www.npmjs.com/package/socket.io)
 ![Downloads](https://img.shields.io/npm/dm/socket.io.svg?style=flat)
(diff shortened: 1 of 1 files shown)


oke11o:
Hi! Gays.

Great PR.

I approve this.

darrachequesne:
I'm not sure we really need yet another badge. Following the bundle size of the client could be interesting though, does packagephobia support this?

styfle:
There is another tool to determine the webpack bundle size of a package called [Bundle Phobia](https://bundlephobia.com/result?p=socket.io)

I would argue that the [Package Phobia](https://packagephobia.now.sh/result?p=socket.io) size is interesting because it shows how durastic the change in dependencies was in v2.1.0

If you bundle your own file, you can use [badge-size](https://github.com/ngryman/badge-size) to display the size of a single file.

All of these options are available on badgen.net like so:

- https://badgen.net/packagephobia/install/socket.io
- https://badgen.net/bundlephobia/minzip/socket.io
- https://badgen.net/badgesize/gzip/https://unpkg.com/socket.io/lib/client.js

styfle:
@darrachequesne Are you aware the `socket.io` version published today jumped from 2.39 MB to 3.21 MB?

https://packagephobia.com/result?p=socket.io

PR #3355, opened 2018-10-23
Throws an error when trying to access the clients of a dynamic namespace

0xClpz:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
Accessing the clients of a dynamic namespace throws because doing io.of(/your-regex/g) returns a namespace with no adapter and the clients methods tries to access namespace.adapter.clients

### New behaviour
Clearer error


darrachequesne:
Thanks!

PR #3456, opened 2019-06-10
Outgoing packet interceptor

mauriceackel:

* [ ] a bug fix
* [X] a new feature
* [ ] other

### Current behaviour
Outgoing packets cannot be intercepted in an easy way (i.e. for checking authentication before sending).

### New behaviour
Interceptors (middleware) can be registered on a namespace and socket in a similar way the connection middleware is registered. When Socket.packet is called to send a packet, all middleware is called, similar to the connection middleware. If no error occurred, the underlying client's packet() method is called. If one middleware returned an error, the error is sent to the client.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -29,6 +29,7 @@
     - [namespace.emit(eventName[, ...args])](#namespaceemiteventname-args)
     - [namespace.clients(callback)](#namespaceclientscallback)
     - [namespace.use(fn)](#namespaceusefn)
+    - [namespace.useinterceptor(fn)](#namespaceuseinterceptorfn)
     - [Event: 'connect'](#event-connect)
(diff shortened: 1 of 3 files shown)


lkho:
Hi all, I came across this idea too. There are many discussions on incoming packet middleware, but none of those mention outgoing packets.

Currently the only way of selecting which socket should receive a broadcast packet is by joining rooms.
However, if the filtering logic is not based on pre-computed data (i.e. can be pre-joined into a specific room), maybe time based or computed with local state variables etc., we do not have any way to hold back the packet once `nsp.emit()` is called. It might be broadcast to multiple nodes by the underlying adapter, and every other nodes will directly write to the engine without ever consulting the namespace.

Why is this closed? I think this will be a useful feature.

PR #3464, opened 2019-07-04
Fix documentation about default origins value. Update API.md

JonaMX:
I was struggling with CORS issues using explicitly the default origins value `*` on development, a value that is retrieved from a `.env` file, but turns out that the default is actually `*:*`.

* [ ] a bug fix
* [ ] a new feature
* [ ] other

### Current documentation
origins (String): the allowed origins (`*`)

### New documentation
origins (String): the allowed origins (`*:*`)


darrachequesne:
@JonaMX thanks!

Issue #3495, opened 2019-10-13
Incorrect query object when connection to different namespace


sebamarynissen:
### Current behaviour

*What is actually happening?*

When one connects to a different namespace with multiplexing **enabled** such as
```js
let one = io.connect('/one', {
  'force new connection': false,
  query: { param: 'one' }
});

let two = io.connect('/two', {
  'force new connection': false,
  query: { param: 'two' }
});
```
the updated query is not parsed correctly on the server. The server side code
```js
let one = io.of('one').on('connection', socket => {
  let { query } = socket.handshake;
  console.log('one', query.param, query.param === 'one');
});
let two = io.of('two').on('connection', socket => {
  let { query } = socket.handshake;
  console.log('two', query.param, query.param === 'two');
});
```
logs
```
one one true
two one false
```
When multiplexing is explicitly disabled by setting 
```js
{ 'force new connection': true }
```
the server logs
```
one one true
two two true
```
as expected.

### Steps to reproduce

Fork https://github.com/sebamarynissen/socket.io-fiddle. Run `npm start`, open up a browser and go to `http://localhost:3000`.

### Expected behaviour

The query object should be the same, regardless of whether multiplexing is enabled or not.

### Setup
- OS: Windows 10
- browser: Chrome 
- socket.io version: 2.3.0

### Other information

When I open up the WS network tab in Chrome devtools, it logs that
```
40/two?param=two
```
is sent to the server, so it seems that at least the client sends the correct query to the server. Hence I guess that this is a server related issue.

sebamarynissen:
I have found the cause of the bug, it is located in socket.js:
```js
Socket.prototype.buildHandshake = function(query){
  var self = this;
  function buildQuery(){
    var requestQuery = url.parse(self.request.url, true).query;
    //if socket-specific query exist, replace query strings in requestQuery
    // The order of query & requestQuery should be swapped! The general
    // `requestQuery` currently overrides socket-specific `query`, but this
    // should be the other way around.
    // return Object.assign({}, query, requestQuery);

    // It should be:
    return Object.assign({}, requestQuery, query);
  }
  return {
    // ...
    query: buildQuery()
  };
};
```

I'm going to file a PR for this.

sebamarynissen:
For people having this issue as well, until #3496 is accepted and published on npm, you can fix it by including the following in your server code:
```js
const url = require('url');
const Socket = require('socket.io/lib/socket');
Socket.prototype.buildHandshake = function(query) {
[...]

PR #3496, opened 2019-10-15
Fix query when multiplexing (#3495)

sebamarynissen:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour

When multiplexing is turned on - which is the default - the query sent to the server when connecting to another namespace is incorrectly processed (as described in #3495).

### New behaviour

The query params are correctly overriden on the server.

See #3495.

### DIFF
--- a/lib/socket.js
+++ b/lib/socket.js
@@ -116,7 +116,7 @@ Socket.prototype.buildHandshake = function(query){
   function buildQuery(){
     var requestQuery = url.parse(self.request.url, true).query;
     //if socket-specific query exist, replace query strings in requestQuery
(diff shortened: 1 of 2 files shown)


sebamarynissen:
I wanted to add a test for this specific case, but I couldn't really figure out how to to do it since the tests don't include the client library.

In any case, all existing tests still pass.

darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/d33a619905a4905c153d4fec337c74da5b533a9e (and included in `socket.io@2.4.0`).

Thanks!

Issue #3567, opened 2020-03-26
socket.client.request.client.getPeerCertificate() always returns null.


Andrewiski:
If client certificates are enabled on the server there is no way to get to the underlying clientCertificate as socket.client.request.client.getPeerCertificate() always returns null.  

This happens even if the the client certificate is valid and authorized.

`socket.client.request.client.authorized = true`

I belive this is the result of a change in node tls functionality.

[https://github.com/nodejs/node/commit/eff8c3e02417652b78436eaa10d049e8c60e5275](https://github.com/nodejs/node/commit/eff8c3e02417652b78436eaa10d049e8c60e5275)

`

var httpsOptions = {
        key: fs.readFileSync(path.join(__dirname, 'serverkey.pem')),
        cert: fs.readFileSync(path.join(__dirname,  'servercert.pem')),
        requestCert: true,
        rejectUnauthorized:false
    };
    https_srv = https.createServer(httpsOptions, app).listen(objOptions.httpsport, function () {
        //console.log('Express server listening on port ' + port);
        writeToLog("info",'Express server listening on https port ' + objOptions.httpsport);
    });
    io.attach(https_srv);

io.on('connection', function (socket) {
let cert = socket.client.request.client.getPeerCertificate();
        if (cert) {
           console.log(cert.subject.CN);
        }else{
           console.log("cert is null");
        }
}
`

Andrewiski:
editing /lib.client.js   and calling getPeerCertificate() on creating allows it to be used later.

function Client(server, conn){
  this.server = server;
  this.conn = conn;
  this.encoder = server.encoder;
  this.decoder = new server.parser.Decoder();
  this.id = conn.id;
  this.request = conn.request;
  **this.cert = conn.request.client.getPeerCertificate();**
  this.setup();
  this.sockets = {};
  this.nsps = {};
  this.connectBuffer = [];
}

socket.client.cert is populated when accessed in io.on('connection', function(socket){
  if( socket.client.cert){
     console.log("this works " + socket.client.cert.subject.CN);
 }
}

darrachequesne:
I could indeed reproduce the issue you're describing, sorry for the delay. I think one should be able to fetch the certificate at the Engine.IO level:

```js
io.engine.on("connection", (rawSocket) => {
  rawSocket.peerCertificate = rawSocket.request.client.getPeerCertificate();
});

io.on("connection", (socket) => {
  console.log(socket.conn.peerCertificate);
});
```

[...]

PR #3568, opened 2020-03-26
Assign the peerCertificate before Node.js closes the connection

Andrewiski:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour
socket.client.request.client.getPeerCertificate() always returns null

### New behaviour
socket.client.peerCertificate is populated before the underlying TLS is closed so it is available 
on socket.on("connection")

I belive this is the result of a change in node tls functionality.

[https://github.com/nodejs/node/commit/eff8c3e02417652b78436eaa10d049e8c60e5275](https://github.com/nodejs/node/commit/eff8c3e02417652b78436eaa10d049e8c60e5275)

[https://github.com/socketio/socket.io/issues/3567](https://github.com/socketio/socket.io/issues/3567)

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -28,6 +28,13 @@ function Client(server, conn){
   this.decoder = new server.parser.Decoder();
   this.id = conn.id;
   this.request = conn.request;
+  if(conn.request.client && conn.request.client.getPeerCertificate){
+    this.peerCertificate = conn.request.client.getPeerCertificate();
+  }else
+  {
+    this.peerCertificate = null;
+  }
+  this.request = conn.request;
   this.setup();
   this.sockets = {};
   this.nsps = {};


dasilvacontin:
This might be hard to merge unless it can easily be tested, or a test is provided.

I also worry whether this takes into account when a session is reused, like discussed in this issue (https://github.com/nodejs/node/issues/3940).

Andrewiski:
Since this is used for client security auth its like having the connection is the certificate is no longer valid the client will need to get a new one and recreate the connection anyways, 

Are you asking that I provide a test?  This is a fix for something that did work but because of a node.js change the getPeerCertificate goes away after initial https socket connection. All we are doing is making it live for the life of the connections.  I suppose if we were trying to save memory like they did we would only make it live during the socket.io connect event then remove it from memory.

if (socket.client.request.client.getPeerCertificate() === null) {
   console.log("Test Passed");
}else{
      console.log("Test Failed");
}

darrachequesne:

Andrewiski:
I assume from the response no one merged my pull request into the main branch so going forward there is no way to view a client certificate on a socket.io connection to determine who the clinet is?  IE want to use TLS client certificates for client auth.  Currently all you know is the cert was signed by the CA not which cert or who is connecting.

darrachequesne:
@Andrewiski that's a valid concern actually, please see my answer here: https://github.com/socketio/socket.io/issues/3567#issuecomment-913089959

PR #3604, opened 2020-05-24
Add failing test for #3555 + the fix itself

### LINKED ISSUE #3555: Served socket.io client does not find client installed in own node_modules
### You want to:

* [x] report a *bug*
* [ ] request a *feature*

### Current behaviour

Sometimes other socket.io-client files are served, pot. very old and so incompatible versions.

### Steps to reproduce (if the current behaviour is a bug)

When it happens that e.g. an other version of socket.io-client is installed in the "root" node_modues and so the socket.io dependencies are installed in it own node_modules the current code logic in https://github.com/socketio/socket.io/blob/master/lib/index.js#L109 will not be able to find it because it only searches "to the top".
It will never find a client installed in node-modules/socket.io/node_modules/socket.io-client/... when there is a client package existing in npm structure „more above“

### Expected behaviour

Find the correct version, so start searching on it's own node_modules. I woud completely switch to use requirer.resolve

### Setup
- OS: irrelevant
- browser: multipe/irrelevant
- socket.io version: 2.3.0

### Other information (e.g. stacktraces, related issues, suggestions how to fix)

Idea for fix: see "Steps to reproduce". 

I could also provide a PR if it has a chance to be adopted.

dasilvacontin:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behaviour

Socket.io server ignores 'socket.io-client' module installed inside `socket.io` when there's an installation of 'socket.io-client' sibling to 'socket.io'.

### New behaviour

Now uses closest 'socket.io-client' installation.

Fix is by @Apollon77, taken from PR #3557. Their fix fixes #3555.

I added a failing test to show it actually had a behavior issue prior to the fix, and give an example of how it fails. Plus to prevent a regression in the future.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -106,7 +106,8 @@ Server.prototype.serveClient = function(v){
   if (!arguments.length) return this._serveClient;
   this._serveClient = v;
   var resolvePath = function(file){
-    var clientPath = require.resolve('socket.io-client'); // resolve the client dep. end in lib/index
+    // resolve the client dep. The resulting string ends with `lib/index.js`
+    var clientPath = require.resolve('socket.io-client');
     return path.normalize(path.join(path.dirname(clientPath), '..', '..', file));
   };
   if (v && !clientSource) {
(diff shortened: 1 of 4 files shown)


darrachequesne:
Superseded by https://github.com/socketio/socket.io/commit/7603da71a535481e3fc60e38b013abf78516d322 (included in `socket.io@3.0.0`). Thanks for the pull request!

PR #3679, opened 2020-11-06
Enables disabling autojoin to a room with the socket's id

bytenik:

* [ ] a bug fix
* [x] a new feature
* [ ] other

### Current behaviour

Upon connection, sockets automatically join a room based on their id.

### New behaviour

Upon connection, sockets only automatically join a room based on their id if the autojoin option is enabled. (It defaults to `true` for backward compatibility.)

Our use case of socket.io doesn't require this "convenience room." However, its existence is a major issue for us because it has allowed a DOS attack -- each room creation is causing allocation of AWS resources via our socket.io adapter, which can't be immediately released. We hit our AWS quota for SNS topic resources and then crash.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -143,6 +143,11 @@ interface ServerOptions extends EngineAttachOptions {
    * @default 45000
    */
   connectTimeout: number;
+  /**
+   * whether or not to automatically join clients to a room bearing the name of the client's id
+   * @default true
+   */
+  autoJoin: boolean;
 }
 
 export class Server extends EventEmitter {
@@ -152,6 +157,8 @@ export class Server extends EventEmitter {
   readonly _parser;
   /** @private */
   readonly encoder: Encoder;
+  /** @private */
+  readonly autoJoin: boolean;
 
   /**
    * @private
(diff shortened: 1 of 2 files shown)


darrachequesne:
@bytenik can't you just ignore the join in your adapter if `id === room`?

bytenik:
@darrachequesne No -- we do sometimes manually want to create that room. Also, I've written the adapter to be general purpose / open source for the good of the community, so I don't want to put application-specific details in there. Also, while I don't recall how the Redis adapter handles rooms, I suspect its probably not great to have the Redis adapter get swarmed with thousands of room creations either.

darrachequesne:
> I suspect its probably not great to have the Redis adapter get swarmed with thousands of room creations either

That used to be the case in the first versions of the Redis adapter, but we had this exact problem (subscribing to a new topic when the socket join a room).

The Redis adapter now subscribes to `socket.io#<namespace>#*` (wildcard) and the broadcast events are published to `socket.io#<namespace>#<room>` and filtered upon reception (https://github.com/socketio/socket.io-redis/blob/5.4.0/index.js#L139-L141).

darrachequesne:

PR #3684, opened 2020-11-08
fix(typings): expose additional needed types

AviVahl:

* [ ] a bug fix
* [ ] a new feature
* [x] other

### Current behaviour

Needed interfaces/types are not exposed. `@types/socket.io` used to export these.

### New behaviour

Exported needed types.

Noticed while upgrading. Will create a similar PR to `socket.io-client`.
May want to consider adding the type modifier to the export, so that the actual abstractions won't be exposed. Downside: it will make the `.d.ts` only compatible with ts@2.8+.

### Off-topic
You using TypeScript and publishing built-in `.d.ts` is awesome.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -712,4 +712,4 @@ emitterMethods.forEach(function(fn) {
 module.exports = (srv?, opts?) => new Server(srv, opts);
 module.exports.Server = Server;
 
-export { Socket }; // for "connect" event typing
+export { Socket, ServerOptions, Namespace };


darrachequesne:
@AviVahl thanks!

ptejada:
Any ideas why I am getting this error when trying to import `Socket`?

image

I am not using typescript and only doing the import for JSDoc type hints. Last version I tried was 3.0.2

PR #3699, opened 2020-11-18
Exports the Namespace and Socket type v3

ptejada:
Exports the Namespace and Socket type from MJS wrapper

* [ ] a bug fix
* [ ] a new feature
* [x] other

### Current behavior
MJS wrapper only exports Server

### New behavior
MJS wrapper exports Server, Namaspace and Socket types 

I would say is a continuation of PR https://github.com/socketio/socket.io/pull/3684 and my comment there
https://github.com/socketio/socket.io/pull/3684#issuecomment-729396811

### DIFF
--- a/wrapper.mjs
+++ b/wrapper.mjs
@@ -1,3 +1,3 @@
 import io from "./dist/index.js";
 
-export const Server = io.Server;
+export const {Server, Namespace, Socket} = io;


darrachequesne:
May I ask what the use case is?

For TypeScript, I understand that it can be used for englobing types:

```ts
import { Namespace } from "socket.io";

class Application {
  private mainNamespace: Namespace;
}
```

But is this needed for ES modules?

ptejada:
@darrachequesne 
Correct is only for ES module and type hinting with JSDoc. IDEs, or at least Jetbrains IDEs cannot resolve the Namaspace or Socket for autocomplete, inline docs and so on. 

Without my update
image

With the update
image

darrachequesne:
@ptejada oh ok. Could you please open an issue in Jetbrains issue tracker then? (here I guess: https://youtrack.jetbrains.com/issues/WEB)

ptejada:
I don't think is an issue of the IDE. What is the purpose of the `wrapper.mjs` file? 

That is the file the IDE is referencing which only exports `Server` at the moment.

I don't have enough information to log bug report for the IDE.  All I know is making the edits in the PR fixes the references.

rheh (approved):

PR #3725, opened 2020-12-11
add more typing info and upgrade prettier

david-fong:
- upgrade prettier to 2.2.0 to gain support for TypeScript's
  new type-only-imports feature.

tests are passing on my local machine (I was a bit confused by the two expected errors at first). You may want to regenerate package-lock.json yourself if you have (what I think is) a healthy level of distrust. I didn't add a malicious package to it, but there's nothing to immediately stop me from doing so.

* [ ] a bug fix
* [ ] a new feature
* [x] other


darrachequesne:
Awesome, thanks!

PR #3744, opened 2021-01-07
add: strict type check in if expressions

rchougule:

* [ ] a bug fix
* [ ] a new feature
* [x] other

### Current behavior
No strict type checking for few cases in `if` expressions

### New behavior
added strict type checks


darrachequesne:
@rchougule thanks!

PR #3760, opened 2021-01-18
Allow explicit creation of namespaces within a dynamic namespace

sebamarynissen:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

I have a use case where I dynamically create namespaces, but when the namespace is created I also set up some middleware. Those middleware functions are not run for the socket that triggered the dynamic namespace to be created though:

```js
sio.of((name, query, next) => {
  if (isValidNamespace(name) {
    next(null, true);
    sio.of(name).use((socket, next) => {

      // This middleware function is not run for the socket that caused the
      // namespace to be created, only for subsequent connections.
      socket.user = authenticate(socket);
      next(null);

    });
  } else {
    return next(null, false};
  }
});
```

### New behavior

The middleware function should be run for the first connection as well.

I am aware that you can register middleware directly on the dynamic namespace like
```js
sio.of((name, query, next) => next(null, true)).use((socket, next) => {
  // Do something
});
```
but in my use case I have to do some setup work when a namespace is created, which requires me to pass the *true* namespace.

I have solved this by manually patching the socket.io server object
```js
const { _checkNamespace } = Server.prototype;
Server.prototype._checkNamespace = function(name, auth, fn) {
  return _checkNamespace.call(this, name, auth, function(nsp) {
    if (nsp) {
      process.nextTick(() => fn.call(this, nsp));
    } else {
      fn(nsp);
    }
  });
};
```
which delays the connection of the socket to the namespace to the next tick, allowing to do some *synchronous* setup work first. It's exactly that solution that I've used in this PR as well. I've added a test case and all existing tests still pass.

A similar solution can be used to backport this to the v2 branch as well as the issue exists here too.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -257,8 +257,11 @@ export class Server extends EventEmitter {
         if (err || !allow) {
           run();
         } else {
-          let nsp = this.parentNsps.get(nextFn.value)!.createChild(name);
-          process.nextTick(() => fn(nsp));
+          let nsp = this._nsps.get(name);
+          if (!nsp) {
+            nsp = this.parentNsps.get(nextFn.value)!.createChild(name);
+          }
+          process.nextTick(() => fn(nsp as Namespace));
         }
       });
     };
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -257,7 +257,8 @@ export class Server extends EventEmitter {
         if (err || !allow) {
           run();
         } else {
-          fn(this.parentNsps.get(nextFn.value)!.createChild(name));
+          let nsp = this.parentNsps.get(nextFn.value)!.createChild(name);
+          process.nextTick(() => fn(nsp));
         }
       });
     };
--- a/test/socket.io.ts
+++ b/test/socket.io.ts
@@ -900,6 +900,27 @@ describe("socket.io", () => {
           });
         });
       });
+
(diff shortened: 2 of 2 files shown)


sebamarynissen:
I've pushed an additional commit which allows both
```js
sio.use((name, query, next) => {
  next(null, true);
  sio.of(name);
});
```
and
```js
sio.use((name, query, next) => {
  sio.of(name);
  next(null, true);
});
```

kohtala:
I just ran into the related problem.

I have application with some objects in database that can emit events. I create a dynamic namespace for the objects and set middleware on the dynamic namespace to see if the user is authorized to access the object and the connect event to set up for receiving some events from the client.

The problem comes when there is an event to emit before a client has connected to the namespace. I take the namespace, but unless a client has already connected, the `of(name)` call creates a namespace without the middleware and event handler. Checking for the existence of the nsp before calling `of` is not a good solution, because it would not work in case there are [multiple servers](https://socket.io/docs/v3/broadcasting-events/#With-multiple-Socket-IO-servers) because I don't think it knows if another server has a client connected.

Do you think your problem would be solved, if the `of` looked up the parent namespace before creating the namespace? So `of` would work the same as a CONNECT.

darrachequesne:

PR #3761, opened 2021-01-18
[3.x/bugfix] Custom "generateId()" works only on the handshake

rennokki:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
Currently, the Session IDs cannot be sent to the client after they were generated by the server using custom generation method. It seems like the initial handshake results with a good ID, but the one generated after `on('connect', ...)` are not custom generated. Here is the connections in the WS tab:

| Data | Length |
| - | - |
| 0{"sid":"5533569206.4606719705","upgrades":[],"pingInterval":25000,"pingTimeout":5000} | 86 |
| 40/echo-app,{"headers":{"X-CSRF-TOKEN":"9Os3T9FdJMaJBLbG15dOG9IWX3XLSGu5uYilVrnI"}} | 83 |
| 40/echo-app,{"sid":"o4IkedYULu37hAzfAAAD"} | 42 |
| 42/echo-app,["subscribe",{"channel":"presence-room","auth":{"headers":{"X-CSRF-TOKEN":"9Os3T9FdJMaJBLbG15dOG9IWX3XLSGu5uYilVrnI"}}}] | 132 |

As you can see, the first `sid` is correct (as it is generated in a custom manner), but the subsequent `sid` is not generated using the custom function.

### New behavior
Not sure if this is a good implementation the way I did it, but I expect this function to be replaceable like the Engine.IO has the `generateId()` function [as explained in the documentation](https://socket.io/docs/v3/server-api/#server-engine-generateId).

### DIFF
--- a/lib/socket.ts
+++ b/lib/socket.ts
@@ -109,11 +109,21 @@ export class Socket extends EventEmitter {
       // @ts-ignore
       this.id = nsp.name !== "/" ? nsp.name + "#" + client.id : client.id;
     } else {
-      this.id = base64id.generateId(); // don't reuse the Engine.IO id because it's sensitive information
+      this.id = this.generateId(); // don't reuse the Engine.IO id because it's sensitive information
     }
     this.connected = true;
     this.disconnected = false;
     this.handshake = this.buildHandshake(auth);
+  }
+
+  /**
+   * Generate an ID for the client.
+   *
+   * @return {any}
+   * @private
+   */
+  private generateId(): any {
+    return base64id.generateId();
   }
 
   /**
--- a/lib/socket.ts
+++ b/lib/socket.ts
@@ -120,9 +120,8 @@ export class Socket extends EventEmitter {
    * Generate an ID for the client.
    *
    * @return {any}
-   * @private
    */
-  private generateId(): any {
+  generateId(): any {
     return base64id.generateId();
   }
 


darrachequesne:
Hmm, it seems the documentation is not up-to-date.

The `generateId()` method is used for the Engine.IO session ID (the one which is found in the query parameters of each request).

It was used as the Socket.IO session ID in v2, but this is indeed not the case anymore: https://github.com/socketio/socket.io/commit/2875d2cfdfa463e64cb520099749f543bbc4eb15

In fact, I think you should be able to override the Socket.IO ID in a namespace middleware:

```js
io.use((socket, next) => {
  socket.id = "my-unique-id";
  next();
});
```

This should work too, provided that the id is unique. Could you please share your use case?

rennokki:
Your workaround worked flawlessly. I was trying that upon a dynamic namespace connection, it should write the socket ID in a custom format for the users. So after using [your workaround](https://github.com/renoki-co/echo-server/blob/1.8.0/src/echo-server.ts#L240), it worked perfectly.

Julien-Marcou:
> 
> 
> In fact, I think you should be able to override the Socket.IO ID in a namespace middleware:
> 
> ```js
> io.use((socket, next) => {
>   socket.id = "my-unique-id";
>   next();
> });
> ```
> 
> This should work too, provided that the id is unique. Could you please share your use case?

@darrachequesne, You can't do this anymore since the v3, with TypeScript, the socket's `id` is a readonly property.

darrachequesne:
@Julien-Marcou that's right! Actually, I'm not sure what the use case for dynamic Socket `id` is, as the value must be unique across all your Socket.IO servers, so you can't use an user ID or the value of a cookie.

Could you please share yours?

Julien-Marcou:
Well, actually I managed to do what I wanted without having to change the socket's id, it required a bit of thinking and sending one more event from the server to the client ^^

Here is my use case :

I'm creating a multiplayer card game with ephemeral session (no database, no account creation), in this game, each player has a dedicated place around the gaming table, so the players must be identified by an id.

If a client gets disconnected (accidentally closing the tab, browser crash, connection lost, even just hitting the F5 key by accident or by reflex), it would soft lock the game, as the user would not be able to regain its place back when they rejoin the game with a new socket.

So, what I wanted to do to allow the user to regain its place back in the game in case of such problem, is create a secret token (stored in a cookie) for each user, then, when I got a new socket connection, if the token sent by the client matches a player in a game, I can link back the player to the new socket and the game continues.

Just for convenience, the players were identified by the id of their client's socket, so if a client gets disconnected and join back using its secret token, then the player's id would not match the socket's id anymore.

To solve my problem I created an abstraction layer around my Player enity, so their id no longer reflect their socket's id, then I notify each client with their own player id (instead of using the socket's id that is directly available on the client side).

PR #3776, opened 2021-01-27
Add implied void return-type and switch to fluent this return type

david-fong:

* [ ] a bug fix
* [ ] a new feature
* [x] other

### Current behavior

- Fluent method signatures return the type of the class.
- Method return types are inferred especially when void.
- Overloads do not specify a general-case type signature.

### New behavior

- Fluent method signatures return the type of the class, or the type of a derived class calling the base method.
- Method return types are explicitly specified. I think this is an improvement in that it explicitly specifies intention, and the compiler will check that the function body doesn't violate that intention. Let me know if you would like me to revert this change.
- General-case overload type signatures are specified. This prevents undesirable behaviour when using conditional types such as `ReturnType`.  [See this StackOverflow answer](https://stackoverflow.com/a/52760599/11107541).

Packages that use TypeScript and depend on this package should not experience any breakage or new compiler warnings as a result of this change.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#polymorphic-this-types

I am motivated to make this change because I am working on a type-utility for socket.io that would enable stronger type checking for `on` and `emit` on both the client and server sides. It involves making interfaces that extend the `Socket` class. Without polymorphic `this`, I wouldn't be able to chain fluent methods such as `on` while keeping my extension-interface's special typings behaviour. Here is a discussion thread I started about this idea: #3777.

### DIFF
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -174,7 +174,7 @@ export class Client {
    * @param {Object} opts
    * @private
    */
-  _packet(packet: Packet, opts?: any) {
+  _packet(packet: Packet, opts?: any): void {
     opts = opts || {};
     const self = this;
 
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -77,7 +77,7 @@ export class Client {
    * @param {Object} auth - the auth parameters
    * @private
    */
-  private connect(name: string, auth: object = {}) {
+  private connect(name: string, auth: object = {}): void {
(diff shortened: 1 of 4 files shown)


darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/9e8f288ca9f14f91064b8d3cce5946f7d23d407c. Thanks a lot!

PR #3787, opened 2021-02-01
fix: init options at addChatMessage

adantoscano:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
Clients do not render participants messages due addChatMessage is trying to access to un initiated options to set fade to false

### New behavior
Now renders


darrachequesne:
Thanks!

PR #3795, opened 2021-02-06
Adapter should not use `typeof`

MickL:
Not sure why `typeof` was used but Adapter is a class and does not need a typeof. The problem is that it was not possible to have a function thats return type is an Adapter:

```
const createAdapter = (): Adapter => { ... };
server.adapter(createAdapter());
```

If there is a reason for use of `typeof` then the types should be `Adapter | typeof Adapter`

* [x] a bug fix
* [ ] a new feature
* [ ] other


MickL:
This PR might be wrong from me. But I create an issue and hope for an answer from the maintainer(s) https://github.com/socketio/socket.io/issues/3796

PR #3822, opened 2021-03-01
Add strictly typed events

### LINKED ISSUE #3742: send functions should have generic types
**Is your feature request related to a problem? Please describe.**

Currently functions like emit are not typable:

```
socket.emit('my-event-name', dto);
```

This applies to both socket.io server and client.

**Describe the solution you'd like**
I would like to have generic types on all emit functions to (optionally) type the send-object and the response-object:

```
socket.emit<void, MyDtoInterface>('my-event-name', dto);
socket.emit<MyResponseDtoInterface, MyDtoInterface>('my-event-name', dto, responseDto => {
   // ...
});
```

Now if an interface changes TypeScript throws an error if the objects are not updated.

**Suggested solution**
Functions could look something like this:

```
// Server:
emit<TInput = any>(ev: string, data: TInput);

// Client:
emit<TResult = any, TInput = any>(ev: string, data: TInput, (response: TResult) => {});
```

MaximeKjaer:

* [ ] a bug fix
* [x] a new feature
* [ ] other

#### User events

`EventEmitter` functions, namely `on` and `emit`, accept a string as the event name, and `any[]` as the remaining parameters. This allows users to:

- Register listeners that listen to wrong event names
- Register listeners that don't match the emitted values

#### Reserved events

Because the event emitter functions aren't strictly typed, it also means that `server.on("connection", (socket: Socket) => { ... })` needs an explicit cast of `socket: Socket`.

#### User events

TypeScript users can define interfaces for the API between server and client by defining interfaces:

```ts
interface Messages {
  hello: string;
}
```

When creating the `Server`, users can optionally pass this interface as a type parameter:

```ts
const server = new Server<Messages>(httpServer);
```

With this in place, calls to `Socket.emit` and `Socket.on` are strictly typed to only allow emitting and listening to `"hello"` events:

```ts
server.on("connection", (socket) => {
  socket.on("hello", (message) => console.log(message)); // works
  socket.on("goodbye", (message) => console.log(message)); // doesn't type check
});
```

It's also possible to define different messages for each direction of the connection:

```ts
interface ClientToServerMessages {
  ping: void;
}

interface ServerToClientMessages {
  pong: void;
}

const server = new Server<ClientToServerMessages, ServerToClientMessages>(httpServer);
server.on("connection", (socket) => {
  socket.on("ping", () => socket.emit("pong")); // works
  socket.on("pong", () => socket.emit("ping")); // doesn't type check
});
```

Note that if no type parameter is passed to `Server`, then events are typed as before.

#### Reserved events

`server.on("connection", socket => { ... })` now correctly infers the `Socket` type for `socket`. No cast needed! Other reserved events are now also correctly inferred. Strict typing of reserved events is also present when no type parameters are passed to `Server`.

This does make this a breaking change if users rely on the `on` callback arguments being of type `any`. This was the case in this project's tests, where `socket` being of typed `any` was used to access private members of `Socket`. The fix is to cast `as any`, which makes the implicit `any` type explicit.

Fixes #3742

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,3 +24,4 @@ jobs:
       - run: npm test
         env:
           CI: true
+      - run: npm run test:types
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,4 +24,3 @@ jobs:
       - run: npm test
         env:
           CI: true
-      - run: npm run test:types
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -17,6 +17,12 @@ import { ServerReservedEventsMap, Socket } from "./socket";
 import type { CookieSerializeOptions } from "cookie";
 import type { CorsOptions } from "cors";
 import type { BroadcastOperator, RemoteSocket } from "./broadcast-operator";
+import {
+  EventsMap,
+  DefaultEventsMap,
+  EventParams,
+  StrictEventEmitter,
+} from "./typed-events";
 
 const debug = debugModule("socket.io:server");
 
@@ -156,57 +162,14 @@ interface ServerOptions extends EngineAttachOptions {
   connectTimeout: number;
 }
 
-/**
- * An events map is an interface that maps event names to their values. Values
- * can be either:
- *
- * - `void`, indicating that the `on` listener takes no parameters
- * - a function type, representing the the type of the `on` listener
- * - any other type, representing the type of the first parameter of the `on` listener
- */
-export interface EventsMap {
-  [event: string]: any;
-}
-
-/**
- * The default events map, used if no EventsMap is given. Using this EventsMap
- * is equivalent to accepting all event names, and any data.
- */
-export interface DefaultEventsMap {
-  [event: string]: (...args: any[]) => void;
-}
-
-/**
- * Returns a union type containing all the keys of an event map.
- */
-export type EventNames<Map extends EventsMap> = keyof Map & (string | symbol);
-
-/** The tuple type representing the parameters of an event listener */
-export type EventParams<
-  Map extends EventsMap,
-  Ev extends keyof Map
-> = Map[Ev] extends (...args: any[]) => void
-  ? Parameters<Map[Ev]>
-  : Map[Ev] extends void
-  ? []
-  : [Map[Ev]];
-
-/**
- * The type of listener callback that listens to the `Ev` event, as defined in `Map`
- */
-export type Listener<
-  Map extends EventsMap,
-  Ev extends keyof Map
-> = Map[Ev] extends (...args: any[]) => void
-  ? Map[Ev]
-  : Map[Ev] extends void
-  ? (...args: []) => void
-  : (...args: [Map[Ev]]) => void;
-
 export class Server<
   UserEvents extends EventsMap = DefaultEventsMap,
   UserEmitEvents extends EventsMap = UserEvents
-> extends EventEmitter {
+> extends StrictEventEmitter<
+  UserEvents,
+  UserEmitEvents,
+  ServerReservedEventsMap<UserEvents, UserEmitEvents>
+> {
   public readonly sockets: Namespace<UserEvents, UserEmitEvents>;
 
   /** @private */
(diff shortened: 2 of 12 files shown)


lib/index.ts
```diff
@@ -645,14 +698,28 @@ export class Server extends EventEmitter {
     return this;
   }
 
+  public on<
```
MaximeKjaer, line 701: I suppose we would also need to strongly type `once`, `removeListener`, etc. Perhaps it might make sense to introduce an `abstract class StrictEventEmitter` between `EventEmitter` and `Server`; this would just override all methods to do all the strict typing, and call `super`'s implementations. What do you think of that approach?

test/socket.io.test-d.ts
```diff
@@ -0,0 +1,211 @@
+"use strict";
+import { DefaultEventsMap, Server, Socket } from "..";
+import { createServer } from "http";
+import { expectError, expectType } from "tsd";
+import "./support/util";
+
+describe("server", () => {
+  describe("no event map", () => {
+    describe("on", () => {
```
MaximeKjaer, line 9: I wrote this test in the same style as Mocha unit tests (`describe` / `it`), but these are not actually run. Perhaps it could make sense to refactor to some other structure, which might be less confusing.

darrachequesne:
@MaximeKjaer I see you've already rebased your PR against the latest changes, nice! :+1:

I'll take a look at it as soon as possible, but it looks good at first sight :ok_hand:

lib/index.ts
```diff
@@ -156,8 +156,58 @@ interface ServerOptions extends EngineAttachOptions {
   connectTimeout: number;
 }
 
-export class Server extends EventEmitter {
-  public readonly sockets: Namespace;
+/**
```
darrachequesne, line 159: How about moving those declarations in their own file? (completely cosmetic, it's your call :angel: )

lib/index.ts
MaximeKjaer: I ended up moving it all to `StrictEventEmitter` to avoid duplicating all the overrides of `on` and `once`. And I added `once`, which has the exact same signature.

test/socket.io.test-d.ts
MaximeKjaer: Good idea! Done :)

.github/workflows/ci.yml
```diff
@@ -24,3 +24,4 @@ jobs:
       - run: npm test
         env:
           CI: true
+      - run: npm run test:types
```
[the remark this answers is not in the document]
MaximeKjaer: Perfect! I placed it under `npm test`. I split that command into two subcommands, just to make it a little more readable.

lib/index.ts
MaximeKjaer: I'm all for it! With the addition of `StrictEventEmitter` I think it made a lot of sense to move it all to a separate file.

test/socket.io.ts
```diff
@@ -952,8 +955,7 @@ describe("socket.io", () => {
       const sio = new Server(srv);
       srv.listen(() => {
         const clientSocket = client(srv, { reconnection: false });
-        clientSocket.on("connect", function init() {
-          clientSocket.removeListener("connect", init);
+        clientSocket.once("connect", () => {
```
MaximeKjaer, line 958: Improving the typing of client `Socket`s showed that `clientSocket.removeListener` does not exist in [component-emitter](https://www.npmjs.com/package/component-emitter), so this had to be fixed in order to get tests to compile.

darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/0107510ba8a0f148c78029d8be8919b350feb633.

I made a few minor updates:

- `<ListenEvents, EmitEvents>` instead of `<UserEvents, UserEmitEvents>`
- `BroadcastOperator<EmitEvents>` instead of `BroadcastOperator<ListenEvents, EmitEvents>`
- `NamespaceReservedEventsMap` instead of `ServerReservedEventsMap`
- `EventParams<...> = Parameters<Map[Ev]>` (so the typings are consistent)

Hope that's OK for you :angel: 

Anyway, awesome work, thanks a lot :heart:

PR #3834, opened 2021-03-12
Fix #3833: Add fallback to untyped event listener

### LINKED ISSUE #3833: Issues with io.on() typings
**Describe the bug**
After upgrading to socket.io 4.0.0, my code no longer compiles when using string enums for event names.

**To Reproduce**

Socket.IO server version: `4.0.0`
TypeScript@4.2.3

*Server*

```js

const io = new Server(3000, {});

// this works fine
io.on("connection", (socket) => {});

enum Event {
    CONNECTION = "connection",
}
// Fails to compile!
// Argument of type '(socket:Socket) => void' is not assignable to parameter of type 'never'.
// ts(2345)
io.on(Event.CONNECTION, (socket) => {});
```

*Client*
N/A

**Expected behavior**
I expect my code to compile like it does for version 2.x and 3.x.

**Platform:**
N/A

MaximeKjaer:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

Due to a compiler bug, we had the following issue (see #3833):

```ts

const io = new Server(3000, {});

// this works fine
io.on("connection", (socket) => {});

enum Event {
    CONNECTION = "connection",
}
// Fails to compile!
// Argument of type '(socket:Socket) => void' is not assignable to parameter of type 'never'.
// ts(2345)
io.on(Event.CONNECTION, (socket) => {});
```

### New behavior

The above code now compiles. However, in the line `io.on(Event.CONNECTION, (socket) => {});`, the lambda parameter `socket` is inferred to be of type `any` instead of `Socket` if an enum member is used. 

Note that this only affects the situation where an enum member is used: If a string literal is used, the correct type (`Socket`) is still inferred.

Also note that inferring `any` in this situation is what was already happening in 3.x. So we are essentially reverting to that, at least until the compiler bug is fixed.

Fixes #3833

### DIFF
--- a/lib/typed-events.ts
+++ b/lib/typed-events.ts
@@ -43,11 +43,23 @@ export type ReservedOrUserListener<
   ReservedEvents extends EventsMap,
   UserEvents extends EventsMap,
   Ev extends ReservedOrUserEventNames<ReservedEvents, UserEvents>
-> = Ev extends EventNames<ReservedEvents>
-  ? ReservedEvents[Ev]
-  : Ev extends EventNames<UserEvents>
-  ? UserEvents[Ev]
-  : never;
+> = FallbackToUntypedListener<
+  Ev extends EventNames<ReservedEvents>
+    ? ReservedEvents[Ev]
+    : Ev extends EventNames<UserEvents>
+    ? UserEvents[Ev]
+    : never
+>;
+
+/**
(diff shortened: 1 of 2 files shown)


westy92:
@darrachequesne can you please take a look at this? This fixes a regression in 4.0.0 that is preventing me from updating.

PR #3843, opened 2021-03-17
update type returned from emit

solomon23:
Using a `true` here breaks using union types to do emits.. ie this no longer works
`(channel ? io.to(channel) : io).emit("stuff", message);`

* [x] a bug fix
* [ ] a new feature
* [ ] other


solomon23:
maybe someone can run with this - i don't know why the unit tests are failing but you shouldn't use true as a type

darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/1a72ae4fe27a14cf60916f991a2c94da91d9e54a. Thanks a lot! :+1:

PR #3855, opened 2021-03-23
Remove unnecessary type annotations in examples

MaximeKjaer:

* [ ] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

`socket` is inferred as type `Socket` as of 4.0.0; it was previously inferred as `any`. The type annotation is no longer needed to cast the `any` to `Socket`.

### New behavior

The `Socket` type annotation can be safely removed. The behavior is unchanged, but it's less code.

Typed events in socket.io 4.0 remove the need for writing type annotations in callbacks of reserved events.

### DIFF
--- a/examples/angular-todomvc/server.ts
+++ b/examples/angular-todomvc/server.ts
@@ -1,4 +1,4 @@
-import { Server, Socket } from "socket.io";
+import { Server } from "socket.io";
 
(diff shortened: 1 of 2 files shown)


darrachequesne:
Good catch, thanks!

PR #3892, opened 2021-04-21
[4.x] Allow RemoteSocket to emit to current rooms

### LINKED ISSUE #3857: v4  RemoteSocket unsupport to/in function
**Is your feature request related to a problem? Please describe.**
v4  RemoteSocket unsupport to/in function
send to everyone in the room except sender when use the controller(KOA or exressjs) ,

**Describe the solution you'd like**
RemoteSocket support to/in function .

rennokki:

* [ ] a bug fix
* [x] a new feature
* [ ] other

### Current behavior

`RemoteSocket` objects cannot emit to other members in some rooms. This simple fix is letting users specify to which rooms the `Socket` (retrieved as `RemoteSocket`) is into, and broadcast to them, excepting the original sender.

```js
io.of('/some-namespace').fetchSockets().then(sockets => {
    sockets.forEach(socket => {
        // Issue: What if you want to emit to everyone but a RemoteSocket?
        if (typeof socket.to !== 'undefined') {
            socket.to('some-room').emit('event', { foo: 'bar' });
        }
    });
});
```

### New behavior

You can now call `.to()` and `.in()` on `RemoteSocket` objects. This way, you can use for example the Redis adapter and use the `RemoteSocket` as a normal `Socket` object when broadcasting to everyone except the sender:

```js
io.of('/some-namespace').fetchSockets().then(sockets => {
    sockets.forEach(socket => {
        socket.to('some-room').emit('event', { foo: 'bar' });
    });
});
```

Closes #3857

In #3857 is described in simple terms the desired behavior.

This also fixes a use case I confront with (and perhaps many will confront).

Consider having a multi-node or a multi-process deployment of Socket.IO, each one exposing a REST API to accept broadcasting messages (same as Pusher API) and their WS server, used by the clients. Both use Redis adapter and are set up properly for high scalability.

If Alice is on Node 1, and Bob is on Node 2, and Alice makes an HTTP request to Node 2 to emit to everyone but to Alice (by using the Client Socket ID), the `fetchSockets()` call will retrieve Bob as `Socket` and Alice as `RemoteSocket`, but calling `.emit()` from `RemoteSocket` will not allow Alice to emit to anyone else.

### DIFF
--- a/lib/broadcast-operator.ts
+++ b/lib/broadcast-operator.ts
@@ -273,7 +273,7 @@ export class RemoteSocket<EmitEvents extends EventsMap>
     this.handshake = details.handshake;
     this.rooms = new Set(details.rooms);
     this.data = details.data;
-    this.operator = new BroadcastOperator(adapter, new Set(details.rooms));
+    this.operator = new BroadcastOperator(adapter, new Set([this.id]));
   }
 
   public emit<Ev extends EventNames<EmitEvents>>(
--- a/lib/broadcast-operator.ts
+++ b/lib/broadcast-operator.ts
@@ -257,20 +257,23 @@ interface SocketDetails {
  * Expose of subset of the attributes and methods of the Socket class
  */
 export class RemoteSocket<EmitEvents extends EventsMap>
-  implements TypedEventBroadcaster<EmitEvents> {
+implements TypedEventBroadcaster<EmitEvents> {
+  public readonly adapter: Adapter;
   public readonly id: SocketId;
   public readonly handshake: Handshake;
   public readonly rooms: Set<Room>;
   public readonly data: any;
 
   private readonly operator: BroadcastOperator<EmitEvents>;
+  private flags: BroadcastFlags = {};
 
   constructor(adapter: Adapter, details: SocketDetails) {
+    this.adapter = adapter;
     this.id = details.id;
     this.handshake = details.handshake;
     this.rooms = new Set(details.rooms);
     this.data = details.data;
-    this.operator = new BroadcastOperator(adapter, new Set([this.id]));
+    this.operator = new BroadcastOperator(adapter, new Set(details.rooms));
   }
 
   public emit<Ev extends EventNames<EmitEvents>>(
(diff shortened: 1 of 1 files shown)


lib/broadcast-operator.ts
```diff
@@ -258,19 +258,22 @@ interface SocketDetails {
...
+  private flags: BroadcastFlags = {};
 
   constructor(adapter: Adapter, details: SocketDetails) {
+    this.adapter = adapter;
     this.id = details.id;
     this.handshake = details.handshake;
     this.rooms = new Set(details.rooms);
     this.data = details.data;
-    this.operator = new BroadcastOperator(adapter, new Set([this.id]));
+    this.operator = new BroadcastOperator(adapter, new Set(details.rooms));
```
darrachequesne, line 276: That does not seem correct. Could you please check?

lib/broadcast-operator.ts
```diff
@@ -280,6 +283,28 @@ export class RemoteSocket<EmitEvents extends EventsMap>
...
   }
 
+  /**
+   * Targets a room when broadcasting.
+   *
+   * @param room
+   * @return self
+   * @public
+   */
+  public to(room: Room | Room[]): BroadcastOperator<EmitEvents> {
```
darrachequesne, line 293: How about creating another operator

```diff
- this.operator = new BroadcastOperator(adapter, new Set([this.id]));
+ this.selfOperator = new BroadcastOperator(adapter, new Set([this.id]));
+ this.broadcastOperator = new BroadcastOperator(adapter, new Set(), new Set([this.id]));
```

`RemoteSocket.to(...)` would simply become `this.broadcastOperator.to(...)`

What do you think?

Note: while we're at it, how about adding the `local`, `broadcast` and `except()` modifiers?

lib/broadcast-operator.ts
rennokki: @darrachequesne Initially I thought that having the rooms set to the RemoteSocket rooms is going to be more logical, but now I see Socket has the same thing. I'll revert it.

lib/broadcast-operator.ts
rennokki: @darrachequesne Would extending the Socket class be a better idea?

darrachequesne: @rennokki yes, why not! Wouldn't the Socket class extend the RemoteSocket class though?

rennokki: @darrachequesne Shouldn't RemoteSocket be a Socket with extra capabilities?

darrachequesne: I'm not sure, since `RemoteSocket` is not an EventEmitter (no `socket.on(...)`) and is missing the `request` attribute, for example.

But please update the PR with what you have in mind, I might be missing something.

darrachequesne:

PR #3990, opened 2021-06-25
Update Github Action

tiziodcaio:

* [ ] a bug fix
* [ ] a new feature
* [X] a code change that improves performance
* [ ] other

### Current behavior

A non-sense expression on cron field is present, node 16 not present

### New behavior

A non-sense expression on cron field removed, node 16 added

Re-add possibly the cron schedule if you want

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,6 @@ name: CI
 on:
   push:
   pull_request:
-  schedule:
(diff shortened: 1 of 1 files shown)


darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/b833f918c8e30cd4fd42b8315325e0a9d6b10caf. I've restored the cron expression, so the CI is periodically run.

Thanks!

PR #4054, opened 2021-08-12
fix: Fixed default export not being typed

GCastilho:
Fixed the default export (exported using 'module.exports') not having type definitions

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
The default export exported using `module.exports` is not exported using `export default`. As a consequence, the default export is not typed and when using typescript as
```typescript
import io from 'socket.io'
io(3000) // This expression is not callable.
```

### New behavior
```typescript
import io from 'socket.io'
io(3000) // Works normally
```

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -822,6 +822,14 @@ export class Server<
   }
 }
 
+export default function io<
+  ListenEvents extends EventsMap = DefaultEventsMap,
+  EmitEvents extends EventsMap = ListenEvents,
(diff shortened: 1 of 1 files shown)


darrachequesne:
Hi, thanks for the pull request!

However, the default export is only kept for backward compatiblity with commonjs usage, users should rather use the named export:

```ts

const io = new Server(3000);
```

Please reopen if needed.

GCastilho:
Okay! Thanks

PR #4074, opened 2021-09-07
Mouse movement path and drawing path are different upon scroll/resize.

deepakpixel:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
While drawing, the lines are drawn with respect to the browser window (using clientX. clientY). When the canvas is more or less than the browser window line drawn are not accurate and are far away from mouse movement. 

Lets make canvas move down by 200px by adding css, canvas{margin-top:200px}
now the line drawn and the mouse movement will be 200px apart.

### New behavior
To fix this we have to decrease the top offset from current.x and left offset from current.y. So that, even after scrolling the offsets are fixed and drawn line is exactly below the cursor.

NA

### DIFF
--- a/examples/whiteboard/public/main.js
+++ b/examples/whiteboard/public/main.js
@@ -59,19 +59,29 @@
     drawing = true;
     current.x = e.clientX||e.touches[0].clientX;
     current.y = e.clientY||e.touches[0].clientY;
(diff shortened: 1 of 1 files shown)


NaveenMohanty:
asdadasd

PR #4099, opened 2021-09-16
Add name field to server cookie option type

douira:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

Using the `name` field in the `cookie` server option produces a type error since the field isn't in `CookieSerializeOptions` but is added by socket.io (engine.io specifically).

### New behavior

The type error doesn't happen anymore.

See [the code](https://github.com/socketio/engine.io/blob/master/lib/server.js#L355) that references `name` in engine.io.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -107,7 +107,7 @@ interface EngineOptions {
    * might be used for sticky-session. Defaults to not sending any cookie.
    * @default false
    */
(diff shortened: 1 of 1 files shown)


darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/033c5d399a2b985afad32c1e4b0c16d764e248cd. Thanks!

PR #4100, opened 2021-09-17
update wrapper.mjs so it provides the same `default` as for CommonJS users

trusktr:

* [ ] a bug fix
* [x] a new feature
* [ ] other

### Current behavior

When migrating from CommonJS to ESM, the end user no longer had the default option available unless they `import io from 'socket.io/dist/index.js'`.

You can see at the bottom of this file: https://unpkg.com/socket.io@4.2.0/dist/index.js

There is this line:

```js
module.exports = (srv, opts) => new Server(srv, opts);
```

which is a convenience for CommonJS users, but ESM users don't have this option (unless they import `dist/index.js`)

### New behavior

Now they can `import io from 'socket.io'` which is similar to `const io = require('socket.io')` as they had when using CommonJS.

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "test:unit": "nyc mocha --require ts-node/register --reporter spec --slow 200 --bail --timeout 10000 test/socket.io.ts",
     "format:check": "prettier --check \"lib/**/*.ts\" \"test/**/*.ts\"",
     "format:fix": "prettier --write \"lib/**/*.ts\" \"test/**/*.ts\"",
-    "prepack": "npm run compile"
+    "prepare": "npm run compile"
   },
   "dependencies": {
     "@types/cookie": "^0.4.1",
--- a/wrapper.mjs
+++ b/wrapper.mjs
@@ -1,3 +1,4 @@
 import io from "./dist/index.js";
 
+export default io;
 export const {Server, Namespace, Socket} = io;


package.json
```diff
@@ -42,7 +42,7 @@
     "test:unit": "nyc mocha --require ts-node/register --reporter spec --slow 200 --bail --timeout 10000 test/socket.io.ts",
     "format:check": "prettier --check \"lib/**/*.ts\" \"test/**/*.ts\"",
     "format:fix": "prettier --write \"lib/**/*.ts\" \"test/**/*.ts\"",
-    "prepack": "npm run compile"
+    "prepare": "npm run compile"
```
trusktr, line 45: use prepare instead of prepack so that installing socket.io from git with npm works

The downside of this is that yarn doesn't follow npm spec, and yarn needs prepack instead of prepare, so install from git with yarn will no longer work. Either npm works, or yarn works. I think it's a better choice to support the standard tool for npmjs.com.

trusktr:
Note, the types need an update. I'm in a plain JS project, so no issue for me, but TS users will get a type error:

Screenshot from 2021-09-16 23-10-37

Because of this, I'll mark this PR as a draft.

trusktr:
I was wrong above, actually if a user tries to import `socket.io/dist/index.js` this will get an error like this:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './dist/index.js' is not defined by "exports" in /home/trusktr/project/node_modules/socket.io/package.json imported from /home/trusktr/project/src/some-file.js
```

So that option is not even on the table, unless they perform an import using a relative path that includes the `node_modules` folder like this:

```js
import io from '../node_modules/socket.io/dist/index.js'
```

But as you know, this is less reliable because the structure inside `node_modules` is not guaranteed.

Another workaround is the user can easily define this function too, though of course it would be better if it existed symmetrically for both CJS or ESM users:

```js
import {Server} from 'socket.io'

const io = (srv, opts) => new Server(srv, opts);
export default io
```

This has the downside that if `socket.io` changes the implementation of the default CJS `io` export, the end user has to update it as well.

darrachequesne:
Hi! This is deliberate. ESM users should use the named export:

```js
```

Reference: https://socket.io/docs/v4/server-initialization/

PR #4145, opened 2021-10-27
feat: add emit hook

naseemkullah:

* [ ] a bug fix
* [x] a new feature
* [ ] other

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -216,7 +216,7 @@ export class Server<
         if (this._nsps.has(name)) {
           // the namespace was created in the meantime
           debug("dynamic namespace %s already exists", name);
-          return fn(this._nsps.get(name) as Namespace);
+          return fn(this._nsps.get(name) ?? false);
         }
         const namespace = this.parentNsps.get(nextFn.value)!.createChild(name);
         debug("dynamic namespace %s was created", name);
--- a/lib/socket.ts
+++ b/lib/socket.ts
@@ -139,6 +139,10 @@ export class Socket<
   private fns: Array<(event: Event, next: (err?: Error) => void) => void> = [];
   private flags: BroadcastFlags & { timeout?: number } = {};
   private _anyListeners?: Array<(...args: any[]) => void>;
+  private emitHook: <Ev extends EventNames<EmitEvents>>(
+    ev: Ev,
+    ...args: EventParams<EmitEvents, Ev>
+  ) => void;
 
   /**
    * Interface to a `Client` for a given `Namespace`.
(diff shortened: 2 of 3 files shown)


darrachequesne:
My thoughts on this:

- we can either use the existing events from the underlying engine:

```js
  socket.conn.on("packet", ({ type, data }) => {
    // called for each packet received
  });

  socket.conn.on("packetCreate", ({ type, data }) => {
    // called for each packet sent
  });
});
```

Note: the downside of this is that the packets are already encoded, so if you emit with `socket.emit("hello", "world");` the `data` in the "packetCreate" event will be `2["hello","world"]`.

I've added the documentation here: https://socket.io/docs/v4/server-socket-instance/#socketconn

- or go all in and add a middleware for any outgoing packets:

```js
// the method name is up to discussion!
socket.useForOutgoing((packet, next) => {
  next();
});
```

Similar to `socket.use()`, but for outgoing packets: https://socket.io/docs/v4/server-socket-instance/#socket-middlewares

Which means we'll have to call it here too: https://github.com/socketio/socket.io-adapter/blob/75854a75183d30d21a70d3b462c2bba7880d20fa/lib/index.ts#L139

What do you think?

naseemkullah:
> Note: the downside of this is that the packets are already encoded, so if you emit with socket.emit("hello", "world"); the data in the "packetCreate" event will be 2["hello","world"].

Thanks @darrachequesne I was looking into leveraging the engine's `packetCreate` event as suggested, but I believe that unless we have an easy and surefire way to decode that in our application code (maybe engine could export a decode function), it would be better to create an emit hook / outgoing middleware. 

This worked for me but seems brittle:

```ts
  socket.conn.on(
    "packetCreate",
    ({ type, data }: { type: string; data: string | Array<string> }) => {
      if (type === "message") {
        try {
          const [eventName, ...args] = JSON.parse(
            (Array.isArray(data) ? data[0] : data).substring(1)
          );

          console.log({ eventName, args });
        } catch (err) {
          console.error("Could not parse packet data");
        }
      }
    }
  );
  ```
  
  WDYT? can we export a decode function from  engine (to replace the `JSON.parse((Array.isArray(data) ? data[0] : data).substring(1));` with something more robust) or  should we make an emit hook in socket.io?

**Comment by darrachequesne** [comment]
@naseemkullah yes, if you need to log the event name, we should totally create an emit hook.

**Comment by lkho** [comment]
A old related PR: https://github.com/socketio/socket.io/pull/3456

**Comment by darrachequesne** [comment]
This feature was implemented in version [4.5.0](https://github.com/socketio/socket.io/releases/tag/4.5.0):

```js
socket.onAnyOutgoing((event, ...args) => {
  console.log(event);
});
```

This does not allow filtering or modifying the packet though.

@naseemkullah thanks!

PR #4189, opened 2021-11-27
Fix reconnection handling in the chat demo app

mixaildudin:

* [ ] a bug fix
* [ ] a new feature
* [x] other

### Current behavior
The source code of the example app `examples/chat`  implies that reconnetion status (failure or success) is shown to the user.
In fact, it's not, because the reconnection events are listened to incorrectly: you're supposed to listen to the Manager's events like `socket.io.on(...)`, not just `socket.on(...)`.

I only wanted to take a look at how to handle disconnections and do reconnections, but here is what it came down to :)

### New behavior
The example works as it's supposed to, you can now see messages like "attempt to reconnect has failed" in the UI.

Also updated README in the example: it says you need to run `npm ci` to install packages, but since lockfiles are gitignored in the `example` folder, you actually need to run `npm i`

### DIFF
--- a/examples/chat/README.md
+++ b/examples/chat/README.md
@@ -6,7 +6,7 @@ A simple chat demo for Socket.IO
 ## How to use
 
 ```
-$ npm ci
+$ npm i
 $ npm start
 ```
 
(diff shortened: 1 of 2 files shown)


darrachequesne:
Good catch, thanks :+1:

PR #4215, opened 2021-12-15
Export type Event used in use()

shayanYousefi:

* [ ] a bug fix
* [ ] a new feature
* [x] other

### New behavior

Now Event type can be used for writing socket middlewares

### DIFF
--- a/lib/socket.ts
+++ b/lib/socket.ts
@@ -107,7 +107,7 @@ export interface Handshake {
   auth: { [key: string]: any };
 }
 
-type Event = [eventName: string, ...args: any[]];
+export type Event = [eventName: string, ...args: any[]];
 
 export class Socket<
   ListenEvents extends EventsMap = DefaultEventsMap,


darrachequesne:
@shayanYousefi hi! Could you please explain your use case?

shayanYousefi:
@darrachequesne Hi! Certainly.

As it is now, when I try to create a middleware for a Socket instance because Event type is not exported, I have to specify the type of event as ``any`` like this which is not ideal:

````
//server-side
io.on('connect',(socket)=>{
    socket.use(logEventsMiddleware);

    function logEventsMiddleware(event: any, next: (err?: (Error | undefined)) => void) {
          let [eventName, ...args] = event
          console.log(`eventName: ${eventName}`)
          console.log(`event args: ${args}`)
          next()
    }
})

````

But with ``Event`` type exposed, using type ``Event`` instead of ``any`` helps both users and IDE to understand what is the underlining structure of event argument.

darrachequesne:
Well, that makes sense. Merged as https://github.com/socketio/socket.io/commit/fc82e44f731d3735b66805dd29eb33f12cb1932e. Thanks!

PR #4233, opened 2022-01-03
chrore(types): fix `SocketData` is not passed down to `Namespace`

### LINKED ISSUE #4229: `SocketData` type is not passed down to the `Namespace`
`SocketData` type is not passed down to the `Namespace` when we do `io.of()`.
That results in `socket.data` being `any` type:

Example: 
```ts
const io = new Server<{},{},{},{ username: string }>()

io.on('connection', socket => {
   socket.data /// { username?: string }
})

io.of('/nps').on('connection', socket => {
   socket.data /// any
})
```

Source code:
https://github.com/socketio/socket.io/blob/51784d0305fda6695af9e325fd14edbabae4c5f4/lib/index.ts#L577-L582

Should be:
```ts
of(...): Namespace<ListenEvents, EmitEvents, ServerSideEvents, SocketData>;
```

OrkhanAlikhanov:
Fixes #4229

https://github.com/socketio/socket.io/issues/4229#issuecomment-1003607352 will be done in another PR

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -112,11 +112,13 @@ export class Server<
   /**
    * @private
    */
-  _nsps: Map<string, Namespace<ListenEvents, EmitEvents, ServerSideEvents>> =
-    new Map();
+  _nsps: Map<
+    string,
+    Namespace<ListenEvents, EmitEvents, ServerSideEvents, SocketData>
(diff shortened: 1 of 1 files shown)


darrachequesne:
Thanks :+1:

PR #4234, opened 2022-01-03
chore(types): make `RemoteSocket.data` type safe

OrkhanAlikhanov:
Adds https://github.com/socketio/socket.io/issues/4229#issuecomment-1003607352

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -769,7 +769,7 @@ export class Server<
    *
    * @public
    */
-  public fetchSockets(): Promise<RemoteSocket<EmitEvents>[]> {
+  public fetchSockets(): Promise<RemoteSocket<EmitEvents, SocketData>[]> {
     return this.sockets.fetchSockets();
   }
 
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -660,7 +660,7 @@ export class Server<
    * @return self
    * @public
    */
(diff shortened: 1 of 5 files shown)


OrkhanAlikhanov:
Ah formatting. Will fix soon

lib/index.ts
```diff
@@ -769,7 +769,7 @@ export class Server<
    *
    * @public
    */
-  public fetchSockets(): Promise<RemoteSocket<EmitEvents>[]> {
```
darrachequesne, line 772: So we would have:

```js
const sockets = await io.fetchSockets();

// with the BroadcastOperator
const sockets = await io.in("room-1").fetchSockets<SocketData>();
```

Is that right?

OrkhanAlikhanov: Exactly

darrachequesne: In that case, I think it would make more sense to go with your first solution [here](https://github.com/socketio/socket.io/issues/4229#issuecomment-1004182988), so we have the same API with and without operator (like `to()` or `except()`).

```ts
const sockets = await io.fetchSockets();
const sockets = await io.in("room-1").fetchSockets(); // no additional types
```

What do you think?

OrkhanAlikhanov: Sounds good to me. Will update the PR

darrachequesne:
@OrkhanAlikhanov awesome, thanks a lot :+1:

PR #4279, opened 2022-02-13
Socket.IO show last seen time

daltonfury42:

* [ ] a bug fix
* [ ] a new feature
* [ ] other

### DIFF
--- a/examples/private-messaging/package.json
+++ b/examples/private-messaging/package.json
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "core-js": "^3.6.5",
+    "moment": "^2.29.1",
     "socket.io-client": "^3.1.1",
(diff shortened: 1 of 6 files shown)


examples/private-messaging/package.json
```diff
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "core-js": "^3.6.5",
+    "moment": "^2.29.1",
```
RobertFischer, line 12: Even the `moment` folks say that you shouldn't use `moment` anymore.

darrachequesne:
@daltonfury42 thanks for your work on this :+1:

I don't think we will merge this, as the readers of the guide might feel overwhelmed by too much details. Thanks again!

PR #4287, opened 2022-02-20
added missing type declaration to .of

Specyk:

* [x] a bug fix

### Current behavior
socketServer.of(...) method has not ability to pass templates like Socket constructor do 

### New behavior
Enables

I've created, but not accepted yet

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -578,7 +578,7 @@ export class Server<
    * @param fn optional, nsp `connection` ev handler
    * @public
    */
-  public of(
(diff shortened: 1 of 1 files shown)


darrachequesne:
Hi! Could you please explain your use case?

You should be able to make it work with:

```js
const namespace: Namespace<CustomListenEvents, CustomEmitEvents> = io.of("/the-namespace");
```

darrachequesne:
Please reopen if needed.

PR #4347, opened 2022-04-19
docs(examples): update create-react-app example

wdwang09:

* [ ] a bug fix
* [ ] a new feature
* [x] other

### Current behavior

Previous create-react-app example can work normally. However, the `useEffect` hook will work multiple times when a new message arrives.

I add two console log to show the problem (line 14 and 26 in [Gist](https://gist.github.com/wdwang09/931347966471780437a5d0288269a1d7)).

before

Socket.IO callback will change the states, so the React component updates and call `useEffect` again.

### New behavior

I fix this problem by passing a second argument to `useEffect` ([doc](https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects)). When states change, don't run `useEffect` again.

after

Other change: I add the logo and other elements generated by create-react-app.

### DIFF
--- a/examples/create-react-app-example/src/App.js
+++ b/examples/create-react-app-example/src/App.js
@@ -1,4 +1,5 @@
 import React, { useState, useEffect } from 'react';
+import logo from './logo.svg';
 import './App.css';
 import io from 'socket.io-client';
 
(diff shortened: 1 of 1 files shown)


darrachequesne:
@wdwang09 good catch, thanks :+1:

PR #4351, opened 2022-04-25
Add https.Server to accepted types

h110m:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
Currently you can't attach a https server to the socket.io instance. You would first need to cast it to any and then to http.Server if you want to use it

### New behavior
Socket.io should now also accept https.Server

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,4 +1,5 @@
 import http = require("http");
+import https = require("https");
 import { createReadStream } from "fs";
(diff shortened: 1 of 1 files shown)


darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/9b43c9167cff817c60fa29dbda2ef7cd938aff51. Thanks :+1:

PR #4367, opened 2022-05-08
fix(types) : added typing for callbacks to socket.emit and socket.on

Aderinom:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Example data
```
export class EmitEvents implements EventsMap {
    'join': (data: string) => string;
}
export class ListenEvents implements EventsMap {
    'message': (data: string) => number;
}
```
### Current behavior : Emit (and derivates)
```
const sock: Socket<EmitEvents, ListenEvents>;
//typing for sock.emit is:
//sock.emit(ev : "join", data : string);
sock.emit('join', "hello", (res)=>{console.log(res)}); // Leads to error since the callback is not defined in typing
```
### New behavior
```
const sock: Socket<EmitEvents, ListenEvents >;
//typing for sock.emit is:
//sock.emit(ev : "join", data : string, cb?:(response : string)=>void); // Uses the return type of the EmitEvent to type the response in the callback
sock.emit('join', "hello", (res)=>{console.log(res)}); // Compiles
```

### Current behavior : On (and derivates)
```
const sock: Socket<EmitEvents, ListenEvents>;
//typing for sock.on is:
//sock.on(ev : "message", listener : (data: string)=>void);
sock.emit('join', "hello", (data, cb)=>{cb("Welcome")}); // Leads to error since the callback is not defined in typing
```

### New behavior
```
const sock: Socket<EmitEvents, ListenEvents>;
//typing for sock.on is:
//sock.on(ev : "message", listener : (data: string, cb:(response:number)=>void)=>void); // Uses return type of the ListenEvent as Cb type
sock.emit('join', "hello", (data, cb)=>{cb(1)}); // Compiles
```

Could possibly break older type script versions from building the Project.
The fix is using following features which I haven't seen be used in other parts of the code:
    1. named tuples        (Released in TS4.0)
    2. ReturnType  utility (Released in TS2.8)

### DIFF
--- a/lib/broadcast-operator.ts
+++ b/lib/broadcast-operator.ts
@@ -3,7 +3,7 @@ import { Handshake, RESERVED_EVENTS, Socket } from "./socket";
 import { PacketType } from "socket.io-parser";
 import type { Adapter } from "socket.io-adapter";
 import type {
-  EventParams,
+  EventEmitArgs,
   EventNames,
   EventsMap,
   TypedEventBroadcaster,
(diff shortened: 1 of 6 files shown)


Aderinom:
Closed as solution exists with declaring the callbacks manually

PR #4594, opened 2023-01-12
Update docker-compose.yml

abriejenny:
Updated docker compose to be in line with current practices. Added `services` and indented all services under it. Also removed `links` as that is intrinsic.

The previous docker compose file did not work with newer versions of docker compose, giving `(root) Additional property server-john is not allowed` since `services` should be the only root property. Marking it as a bug fix.

* [x] a bug fix
* [ ] a new feature
* [ ] other


darrachequesne:
@abriejenny thanks!

PR #4602, opened 2023-01-19
Add cleanupEmptyChildNamespaces server option

stevebaum23:
- When a socket disconnects from a child namespace (those created by a ParentNamespace), if this option is enabled and there are no other sockets connected, it will call nsp.adapter.close() and remove the namespace.
- The namespace can be connected to later (it will be recreated)
- Added some tests to cover the new code changes

* [ ] a bug fix
* [x] a new feature
* [ ] other

### Current behavior
Currently, child namespaces (those created by a `ParentNamespace`, maybe dynamic namespace is the correct term?) exist forever while the server is running. This can cause issues with certain adapters (like redis) where it continues to subscribe and never unsubscribes, as pointed out in [this  issue](https://github.com/socketio/socket.io-redis-adapter/issues/480).

### New behavior
This PR adds a new server option, `cleanupEmptyChildNamespaces`. With this option enabled, when a client socket disconnects, if there are no other sockets connected  to the namespace, the following with occur:

- Call `nsp.adapter.close()`
- Remove the namespace from the servers internal namespace list (`_nsps`)
- Remove the namespace from it's parent list (`ParentNamespace.children`)

Calling `adapter.close()` will allow the adapters to clean up any subscriptions. 

Removing the namespace from `_nsps` is necessary, otherwise the adapters won't re-subscribe when the namespace is connected to again. 

Removing the namespace from the `ParentNamespace.children` is to prevent memory leaks, since the old namespace is no longer used.

One thing I'm unsure of is how this will affect the other adapters. I only use the redis adapter, which doesn't implement close yet (I will be creating a new PR to implement it). I don't know if calling close on the other adapters would have some bad side effects.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -96,6 +96,11 @@ interface ServerOptions extends EngineOptions, AttachOptions {
      */
     skipMiddlewares?: boolean;
   };
+  /**
+   * whether or not to remove child namespaces that have no sockets connected to them
+   * @default false
+   */
+  cleanupEmptyChildNamespaces: boolean;
 }
 
 /**
@@ -153,6 +158,7 @@ export class Server<
    *
    */
   public engine: BaseServer;
+  public readonly cleanupEmptyChildNamespaces: boolean;
 
   /** @private */
   readonly _parser: typeof parser;
@@ -226,6 +232,7 @@ export class Server<
     this.path(opts.path || "/socket.io");
     this.connectTimeout(opts.connectTimeout || 45000);
     this.serveClient(false !== opts.serveClient);
+    this.cleanupEmptyChildNamespaces = !!opts.cleanupEmptyChildNamespaces;
     this._parser = opts.parser || parser;
     this.encoder = new this._parser.Encoder();
     this.opts = opts;
--- a/lib/namespace.ts
+++ b/lib/namespace.ts
@@ -147,19 +147,24 @@ export class Namespace<
   /** @private */
   _ids: number = 0;
 
+  private cleanupEmptyNamespaceFunc?: (nsp: Namespace) => void;
+
   /**
    * Namespace constructor.
    *
    * @param server instance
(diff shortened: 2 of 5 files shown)


darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/5d9220b69adf73e086c27bbb63a4976b348f7c4c. Thanks a lot :+1:

carera:
Hello @stevebaum23 and @darrachequesne 
We greatly benefited from your leak fix in this PR!

We have stumbled upon a similar issue, when using Redis adapter, tho not sure how should we go about fixing it. It appears that adapter cleanup isn't called when a socket.io middleware throws exception:

- When we call next(new Error('Foo') in a socketIO middleware (as suggested [here](https://socket.io/docs/v3/middlewares/#registering-a-middleware)), socketIO lib calls _cleanup on error [here](https://github.com/socketio/socket.io/blob/main/lib/namespace.ts#L333)
- (note that, at this point, redis-adapter is already subscribed to topics in Redis)
- [_cleanup](https://github.com/socketio/socket.io/blob/main/lib/socket.ts#L773) implementation calls leaveAll, which is [here](https://github.com/socketio/socket.io/blob/main/lib/socket.ts#L589) and which calls adapter.delAll
- socketIO adapter interface defines delAll [here](https://github.com/socketio/socket.io-adapter/blob/main/lib/index.ts#L137) (it calls this._del in a loop)
- [redis adapter](https://github.com/socketio/socket.io-redis-adapter/blob/main/lib/index.ts) however does not override/extend this method with custom behaviour, which, i think, should clean up Redis subscriptions, in case there are no more clients connected to given topic

I attempted a PR [here](https://github.com/socketio/socket.io-redis-adapter/pull/508) but that does not seem to be the correct approach - we should only clean up when there are no more sockets

Would you be able to help?

PR #4675, opened 2023-03-28
fix(client): auth init error

blastZ:
Parser maybe parse `undefined` value to `null`, like [msgpackr](https://github.com/kriszyp/msgpackr#readme)'s `encodeUndefinedAsNil` option.

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
Throw error.

### New behavior
Work fine.

### DIFF
--- a/lib/client.ts
+++ b/lib/client.ts
@@ -115,6 +115,8 @@ export class Client<
    * @private
    */
   private connect(name: string, auth: Record<string, unknown> = {}): void {
+    auth === null && (auth = {});
+    
     if (this.server._nsps.has(name)) {
       debug("connecting to namespace %s", name);
       return this.doConnect(name, auth);


xs-fahim (approved):

darrachequesne:
Hi! Thanks for opening this.

I don't think this should be fixed here though. `null` is not a valid payload, we expect either `undefined` or an object. What do you think?

blastZ:
The value is `undefined`, but some parser will encode and decode it to `null` like [msgpackr](https://github.com/kriszyp/msgpackr#readme)

`encodeUndefinedAsNil` - Encodes a value of undefined as a MessagePack nil, the same as a null.

If someone use this feature, socket.io will throw `TypeError` when read property of `auth`

darrachequesne:
I'm going to close this, as I think this is the duty of the parser to make sure to not return a `null`.

Please reopen if needed.

PR #4701, opened 2023-04-30
docs: fix missing types in basic CRUD example

Jood80:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
I tried pulling the latest version of the [Basic CRUD application with Socket.IO](https://github.com/socketio/socket.io/tree/main/examples/basic-crud-application) and after installing the dependencies and starting both the client and the server apps, I got this error on the client side:
```
error TS2591: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i @types/node` and then add `node` to the types field in your tsconfig.
```

### New behavior
- ~installed `@types/node` as devDependency in the `server` directory.~
- updated `angular-client/tsconfig.app.json` with 
```js
"types": [ "node"]
``` 
PS: the code worked on me without installing `@types/node` dep, however I thought it would be recommended to install it anyway. 
none

### DIFF
--- a/examples/basic-crud-application/angular-client/tsconfig.app.json
+++ b/examples/basic-crud-application/angular-client/tsconfig.app.json
@@ -3,7 +3,9 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "outDir": "./out-tsc/app",
-    "types": []
+    "types": [
+      "node"
+    ]
   },
   "files": [
     "src/main.ts",
--- a/examples/basic-crud-application/server/package.json
+++ b/examples/basic-crud-application/server/package.json
@@ -26,7 +26,6 @@
   "devDependencies": {
     "@types/chai": "^4.2.16",
     "@types/mocha": "^10.0.0",
-    "@types/node": "^18.16.3",
     "@types/uuid": "^8.3.0",
     "chai": "^4.3.4",
     "mocha": "^10.0.0",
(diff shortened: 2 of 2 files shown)


darrachequesne:
Hi! `@types/node` should already be included when installing the `socket.io` package: https://github.com/socketio/engine.io/blob/911d0e35757ea9ee93d1807c401c734661615e96/package.json#L36

```
$ npm ls @types/node
basic-crud-server@0.0.1 /examples/basic-crud-application/server
├─┬ socket.io@4.6.1
│ └─┬ engine.io@6.4.1
│   ├─┬ @types/cors@2.8.13
│   │ └── @types/node@18.16.3 deduped
│   └── @types/node@18.16.3 deduped
└─┬ ts-node@10.9.1
  └── @types/node@18.16.3
```

Isn't that the case for you? Which version of TypeScript are you using?

Jood80:
@darrachequesne Oh! thank you for mentioning that, I didn't know it was there already. That makes sense now, why the error is gone after updating only the `angular-client/tsconfig.app.json` file without the need to install `@types/node` as I mentioned.  
As for my Typescript version it's 4.5.4. Please let me know if the error appears only on my end so I can close the PR.

darrachequesne:
I think I have finally found the culprit! I've extracted the events to the `common` directory, so that the typescript compiler does not try to compile the types from the `server` directory:

```
examples/basic-crud-application/
├── angular-client
├── common
├── server
└── vue-client
```

Related: https://github.com/socketio/socket.io/commit/3848280125f35768e74be4d778d7b55acf82f7df

Issue #4537, opened 2022-11-30
`SocketData` being wrapped with `Partial` can create the need to use non-null assertions


navidmafi:
**Is your feature request related to a problem? Please describe.**

Socket.IO allows to add arbitrary data to Socket instance with `Socket.data` which is very nice and handy. the problem arises when using this data in a type-safe manner.
Please take a look at the example below:

In this example, the middleware ensures that the socket connection is authenticated and if it is, it adds some data to socket instance for later usage.
As you can see, the Server object is typed with ISocketData which looks like this :
```ts
export interface ISocketData {
        userID: string;
        userName: string;
        isEmailVerified: boolean;
        joinedRoomUUID: string | null;
        joinedRoomInviteCode: string | null;
    }
```
![image](https://user-images.githubusercontent.com/46371349/204761185-0b3f649a-2e9f-4906-bca3-728cb4207f54.png)
  
But since Socket.data type is internally wrapped with `Partial`, undefined type will be added to all of the properties.
This can causes the need to use non-null assertion or any type of TypeAssertion when using any of the properties in `Socket.data`, since of course Typescript wouldn't know the logical context.

![image](https://user-images.githubusercontent.com/46371349/204763206-fd2935d1-4d35-473f-8d91-f0495c2fa3aa.png)
  

**Describe the solution you'd like**

Not wrapping SocketData with `Partial` internally. Current codebases will work fine and properties can still have the undefined type by just adding it in type of the value passed to SocketData like below if the business logic or anything requires them to have the `undefined` type.

![image](https://user-images.githubusercontent.com/46371349/204766757-f7aae5d2-efbb-4985-b62d-1300c2974473.png)

**Describe alternatives you've considered**
Using manual assertions introduced in TS 3.7. please check the links below with the first one being the same example said here.

https://twitter.com/NavidMafi/status/1597884685551513600
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

**Additional context**
![image](https://user-images.githubusercontent.com/46371349/204767888-7eed04fe-3586-475d-bffb-478a66883729.png)

![image](https://user-images.githubusercontent.com/46371349/204765040-5b66fc9e-1a0e-4957-8ce5-68689fa87a19.png)

darrachequesne:
[...]

PR #4740, opened 2023-06-18
Update socket.ts: Change the socket.data type from Partial<SocketData…

SyedTayyabUlMazhar:

* [x] a bug fix
* [ ] a new feature
* [x] other

### Current behavior
The type of socket.data is Partial<SocketData>. If I am passing SocketData type as {id: number, name: string}, then as I read socket.data.name I would have to use non-null assertions even though my own type implies that id would always be defined.

### New behavior
Now, if someone wants to specify that a few or all of their properties might not be available at runtime, they can do so by specifying that in their own type.

See [issue 4537](https://github.com/socketio/socket.io/issues/4537)

### DIFF
--- a/lib/socket.ts
+++ b/lib/socket.ts
@@ -202,7 +202,7 @@ export class Socket<
    * Additional information that can be attached to the Socket instance and which will be used in the
    * {@link Server.fetchSockets()} method.
    */
-  public data: SocketData = {};
+  public data: SocketData = {} as SocketData;
   /**
    * Whether the socket is currently connected or not.
    *
--- a/lib/socket.ts
+++ b/lib/socket.ts
@@ -202,7 +202,7 @@ export class Socket<
    * Additional information that can be attached to the Socket instance and which will be used in the
    * {@link Server.fetchSockets()} method.
    */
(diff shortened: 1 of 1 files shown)


darrachequesne:
@SyedTayyabUlMazhar thanks!

Edgarruiz8585:
> ### The kind of change this PR does introduce
> 
> * [x] a bug fix
> * [ ] a new feature
> * [ ] an update to the documentation
> * [ ] a code change that improves performance
> * [x] other
> 
> ### Current behavior
> The type of socket.data is Partial<SocketData>. If I am passing SocketData type as {id: number, name: string}, then as I read socket.data.name I would have to use non-null assertions even though my own type implies that id would always be defined.
> 
> ### New behavior
> Now, if someone wants to specify that a few or all of their properties might not be available at runtime, they can do so by specifying that in their own type.
> 
> 
> ### Other information (e.g. related issues)
> See [issue 4537](https://github.com/socketio/socket.io/issues/4537)
> 
> 
> 

- [x] verificado por Edgarruiz8585 #4740 dentro de Edgarruiz8585

Issue #4772, opened 2023-07-19
Server is not cleaning up namespaces after client is rejected in middleware.


carera:
Hello team! 💜

**Describe the bug**
When using SocketIO with adapter (namely redis adapter in my case) and dynamic namespaces, the namespace isn't cleaned up when SocketIO server throws an error in middleware.

I believe this is a continuation of a fix @stevebaum23 previously made here: https://github.com/socketio/socket.io/pull/4602
The new flag `cleanupEmptyChildNamespaces` is cleaning up namespaces when last client disconnects gracefully. It is however not cleaning up properly when client is rejected in middleware.

**To Reproduce**

Socket.IO server version: `4.7.1`

*Server*

```js
import { Server } from "socket.io";
import Redis from "ioredis";
import { RedisAdapter } from "@socket.io/redis-adapter";

const pubClient = new Redis({
  host: "localhost",
  port: 6370,
});
const subClient = pubClient.duplicate();

function redisAdapterFactory(nsp) {
  return new RedisAdapter(nsp, pubClient, subClient);
}

const io = new Server({
  transports: ["websocket"],
  adapter: redisAdapterFactory,
  cleanupEmptyChildNamespaces: true,
});

const ns = io.of(/\/path.*/);

ns.use((fn, next) => {
  if (Math.random() > 0.5) {
    next(new Error("not allowed!"));
  } else {
    next();
  }
});

ns.on("connection", (socket) => {
  console.log(`connect ${socket.id}`);

  socket.on("disconnect", () => {
    console.log(`disconnect ${socket.id}`);
  });
});

io.listen(3000);
```

Socket.IO client version: `4.7.1`

*Client*

```js
import { io } from "socket.io-client";

const socket = io("ws://localhost:3000/path/abc", {
  transports: ["websocket"],
});

socket.on("connect", () => {
  console.log(`connect ${socket.id}`);
  socket.disconnect();
});

socket.on("disconnect", () => {
  console.log("disconnect");
});

socket.on("error", (err) => {
  console.log(err, "error");
});

socket.on("connect_error", (err) => {
  console.log(err, "connect_error");
});
```

**Expected behavior**

In the server code, I have a middleware that allows the client to connect 50% of the time (at random). The client's only job is to connect and disconnect straight away.
When observed on redis (using e.g. `redis-cli` and `MONITOR` command), when client connects and gracefully disconnects, the Redis namespace is cleaned up (we observe subscribe and unsubscribe events):

```
[...]

PR #4773, opened 2023-07-19
Clean up namespaces after client is rejected in middleware

carera:

* [x] a bug fix
* [ ] a new feature
* [ ] other

Change is fixing bug described here: https://github.com/socketio/socket.io/issues/4772

### DIFF
--- a/lib/socket.ts
+++ b/lib/socket.ts
@@ -758,6 +758,8 @@ export class Socket<
     }
 
     this._cleanup();
+    this.client._remove(this);
+    this.connected = false;
     this.emitReserved("disconnect", reason, description);
     return;
   }
@@ -770,8 +772,6 @@ export class Socket<
   _cleanup() {
     this.leaveAll();
     this.nsp._remove(this);
-    this.client._remove(this);
-    this.connected = false;
     this.join = noop;
   }
 
--- a/lib/socket.ts
+++ b/lib/socket.ts
@@ -758,9 +758,6 @@ export class Socket<
     }
 
     this._cleanup();
-    this.nsp._remove(this);
-    this.client._remove(this);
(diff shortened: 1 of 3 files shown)


lib/socket.ts
```diff
@@ -758,9 +758,6 @@ export class Socket<
     }
 
     this._cleanup();
```
carera, line 760: _cleanup is called here and in two more places: https://github.com/socketio/socket.io/blob/main/lib/namespace.ts#L327-L333

I believe the `this.nsp._remove(this);` should be called in all these cases

lib/socket.ts
```diff
@@ -772,6 +769,9 @@ export class Socket<
    */
   _cleanup() {
     this.leaveAll();
+    this.nsp._remove(this);
+    this.client._remove(this);
+    this.connected = false;
```
darrachequesne, line 774: I don't think `this.client._remove(this);` and `this.connected = false;` are needed here, since the socket is not actually connected if it's rejected by the middleware. What do you think?

carera: I agree, I just thought it does no harm since `client.remove` has a guard check and no-ops if the socket is not actually connected, but yeah, let me move it out

carera: Done.

darrachequesne:
Thanks a lot :+1:

carera:
Thank you for merging. What's the release process of this repo? Do I need to do something for this to roll out?

darrachequesne:
@carera we have a few other bug fixes in progress, I'll keep you updated.

darrachequesne:
Update: released in version [4.7.2](https://github.com/socketio/socket.io/releases/tag/4.7.2) :rocket:

PR #4817, opened 2023-09-02
improve emit types

### LINKED ISSUE #4813: Type issues with `emit` and `emitWithAck`
**Describe the bug**
When using `emit`, `emitWithAck`, and the server side variants, there are issues with the types of Broadcaster, Server, Namespace, and Socket  types.

**To Reproduce**

*Server*

```js

export interface ServerToClientEvents {
  basicEmit: (a: number, b: string) => void;
  withAck: (d: string, callback: (e: number) => void) => void;
  // multiple Ack arguments doesn't work on broadcasts - single target
  withAck2: (d: string, callback: (e: number, f: string) => void) => void;
}
export interface ClientToServerEvents {}
const io = new Server<ClientToServerEvents,ServerToClientEvents>(3000, {});

//Emits
// no callback
io.timeout(1).emit('basicEmit',3,'')
// Likely should be a type error as it doesn't have a callback at all
// Currently just times out...

// One arg in callback
io.emit('withAck', 'hello', (...args) => { });
//=> ❌ typeof args === [number], should be [number[]]
io.timeout(1).emit('withAck', 'hello', (...args) => {});
//=> ✅ typeof args === [Error, number[]]
io.to('1').to('2').emit('withAck','',(...args)=>{})
//=> ❌ typeof args === [number], should be [number[]]

// Two args in callback
io.emit('withAck2', 'hello', (...args) => { });
//=> ❌ typeof args === [number,string], should be [number[]]
io.timeout(1).emit('withAck2','hello',(...args)=>{});
//=> ❌ typeof args === [Error,number,string], should be [Error,number[]]

// EmitWithAcks
// no callback
io.emitWithAck('basicEmit',3,'')
//=> type error: Expected 2 arguments, got 3 (the last argument is clipped off)
// Should just be a type-error, as without `timeout` applied, it will likely just never resolve
// if you apply a timeout, then it just times out eventually due to not having a callback

// One arg in callback
io.emitWithAck('withAck','hello');
//=>❌ type is Promise<any>, should be Promise<number[]>
io.timeout(1).emitWithAck('withAck','hello')
//=>✅ type is Promise<number[]>

// Two args in callback - doesn't even work in timeout broadcaster
io.emitWithAck('withAck2','hello');
//=>❌ type is Promise<any>, should be Promise<number[]>
io.timeout(1).emitWithAck('withAck2','hello')
//=>❌ type is Promise<any>, should be Promise<number[]>


  // emit is fine on the socket instance itself
  socket.broadcast.emit('withAck','hello',(...args)=>{});
  //=>❌ typeof args === [number], should be [number[]]
  socket.broadcast.timeout(1000).emit('withAck', 'hello',(...args)=>{});
  //=>✅ typeof args === [Error, number[]]
  // important to note that whatever types exist needs to work with socket.broadcast.timeout(1) AND socket.timeout(1).broadcast
  socket.timeout(1).broadcast.timeout(1).emit('withAck','hello',(...args)=>{});
  //=>❌ typeof args === [Error, Error, number[]], should be [Error, number[]]
  // Needs to not re-decorate the event if an error was already added. (not sure what benefit allowing this gives in the first place
  
  // emitWithAck is backwards here vs the broadcast based ones...
  socket.emitWithAck('withAck','hello');
  //=>✅ Promise<number>
  socket.timeout(1).emitWithAck('withAck','hello');
  //=>❌ Promise<any>, should be Promise<number>
  
  // two args
  socket.emitWithAck('withAck2','hello');
  //=>❌ Promise<any>, should be Promise<number>
  socket.timeout(1).emitWithAck('withAck2','hello');
  //=>❌ Promise<any>, should be Promise<number>
});
```

Socket.IO client version: `x.y.z`

*Client*
I haven't checked on the client, but I assume the same types are used on here too (to some extent)

**Expected behavior**
Type definitions should be consistent with what socket.io is doing so that we can fully rely on them. As of the moment, they are wrong in a few places that are either frustrating (`any` instead of a real type) or problematic (`number` instead of `number[]`).

**Platform:**
 - Typescript 5.2

**Additional context**
I have some first passes at solving some of these issues that I can work towards improving and turning into a PR, if that's desired.

ZachHaber:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
The types of the `emit`, `emitWithAck`, `serverSideEmit`, and `serverSideEmitWithAck` have issues with their types.

They revert to `any` very often or aren't arrays when they should be (server/namespace emit and server/namespace => broadcastOperator emit).

### New behavior
This fixes most of this behavior and adds tests for them.

Fixes #4813

### DIFF
--- a/lib/namespace.ts
+++ b/lib/namespace.ts
@@ -13,6 +13,7 @@ import {
   DecorateAcknowledgementsWithMultipleResponses,
   FirstNonErrorArg,
   DecorateAcknowledgements,
+  RemoveAcknowledgements,
 } from "./typed-events";
 import type { Client } from "./client";
 import debugModule from "debug";
@@ -119,7 +120,7 @@ export class Namespace<
   SocketData = any
 > extends StrictEventEmitter<
   ServerSideEvents,
-  EmitEvents,
+  RemoveAcknowledgements<EmitEvents>,
   NamespaceReservedEventsMap<
     ListenEvents,
     EmitEvents,
@@ -441,37 +442,12 @@ export class Namespace<
    */
   public emit<Ev extends EventNames<EmitEvents>>(
     ev: Ev,
-    ...args: EventParams<EmitEvents, Ev>
+    ...args: EventParams<RemoveAcknowledgements<EmitEvents>, Ev>
   ): boolean {
     return new BroadcastOperator<EmitEvents, SocketData>(this.adapter).emit(
       ev,
       ...args
     );
-  }
-
-  /**
-   * Emits an event and waits for an acknowledgement from all clients.
-   *
-   * @example
-   * const myNamespace = io.of("/my-namespace");
-   *
-   * try {
-   *   const responses = await myNamespace.timeout(1000).emitWithAck("some-event");
-   *   console.log(responses); // one response per client
-   * } catch (e) {
-   *   // some clients did not acknowledge the event in the given delay
-   * }
-   *
-   * @return a Promise that will be fulfilled when all clients have acknowledged the event
-   */
-  public emitWithAck<Ev extends EventNames<EmitEvents>>(
-    ev: Ev,
-    ...args: AllButLast<EventParams<EmitEvents, Ev>>
-  ): Promise<FirstNonErrorArg<Last<EventParams<EmitEvents, Ev>>>[]> {
-    return new BroadcastOperator<
-      DecorateAcknowledgementsWithMultipleResponses<EmitEvents>,
-      SocketData
-    >(this.adapter).emitWithAck(ev, ...args);
   }
 
   /**
--- a/lib/namespace.ts
+++ b/lib/namespace.ts
@@ -11,7 +11,6 @@ import {
   Last,
   FirstArg,
   DecorateAcknowledgementsWithMultipleResponses,
-  FirstNonErrorArg,
   DecorateAcknowledgements,
   RemoveAcknowledgements,
 } from "./typed-events";
--- a/lib/namespace.ts
+++ b/lib/namespace.ts
@@ -10,7 +10,9 @@ import {
   AllButLast,
   Last,
   FirstArg,
-  SecondArg,
+  DecorateAcknowledgementsWithMultipleResponses,
+  FirstNonErrorArg,
+  DecorateAcknowledgements,
 } from "./typed-events";
 import type { Client } from "./client";
 import debugModule from "debug";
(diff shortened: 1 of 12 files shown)


ZachHaber:
Open question:
- Should the `emitWithAck`/`serverSideEmitWithAck` functions only allow events names that have ack functions?

ZachHaber:
@darrachequesne, I've improved the types further, and also removed `emitWithAck` from the server and namespace.
Some of these type changes will need to be mirrored to the client library to fix up emitWithAck types there. I also can't change the docs website to adjust it to the changes.

I'll be away from computers from Saturday till October, so feel free to make any needed changes to the PR.

ZachHaber:
@darrachequesne, could you review this when you get a chance?

darrachequesne:
@ZachHaber sorry for the delay, I'm taking a look at this.

lib/typed-events.ts
```diff
@@ -241,3 +301,11 @@ export type DecorateAcknowledgementsWithMultipleResponses<E> = {
     ? (...args: ExpectMultipleResponses<Params>) => Result
     : E[K];
 };
+
+export type RemoveAcknowledgements<E> = {
```
darrachequesne, line 305: I'm not sure to understand here, shouldn't the events with an acknowledgement be flltered out?

package.json
```diff
@@ -52,7 +52,8 @@
     "debug": "~4.3.2",
     "engine.io": "~6.5.2",
     "socket.io-adapter": "~2.5.2",
-    "socket.io-parser": "~4.2.4"
+    "socket.io-parser": "~4.2.4",
+    "type-fest": "^3.13.1"
```
darrachequesne, line 56: We try to keep our prod dependencies to the bare minimum, so I think we'll indeed need to copy the types from `type-fest`.

lib/namespace.ts
```diff
@@ -252,7 +255,10 @@ export class Namespace<
    * @return a new {@link BroadcastOperator} instance for chaining
    */
   public to(room: Room | Room[]) {
-    return new BroadcastOperator<EmitEvents, SocketData>(this.adapter).to(room);
+    return new BroadcastOperator<
+      DecorateAcknowledgementsWithMultipleResponses<EmitEvents>,
```
darrachequesne, line 259: Couldn't we move the `DecorateAcknowledgementsWithMultipleResponses` type to the `BroadcastOperator` class, so we don't have to apply it everytime?

ZachHaber: I definitely think that would be nicer; however after looking at the code, I don't think we can.
The `BroadcastOperator` has an `expectSingleResponse` flag, which changes the behavior between single responses and multiple. With that, the options I see are to either: apply the decorations when switching to a `BroadcastOperator`, or to setup the expectSingleResponse as a third generic type parameter on the BroadcastOperator.

darrachequesne:
Awesome work, thanks a lot :+1:

lts20050703:
@ZachHaber Awesome work, thank you for your contribution! However, it seems like your PR caused #4914, is there any chance you can look into this?

ZachHaber:
@lts20050703 I'll take a look and see if I can figure that out!

PR #4853, opened 2023-10-16
fix(typings): accept an Http2Server non-secure as http-server instance

0x0a0d:
http server should accept non-secure http2server instance


darrachequesne:
Hi! Out of curiosity, what is your use case? (since there are no browsers known that support [unencrypted HTTP/2](https://http2.github.io/faq/#does-http2-require-encryption))

0x0a0d:
We can use it not only on browser but also on server to server that don't require TLS due to performance

darrachequesne:
@0x0a0d thanks :+1:

PR #4878, opened 2023-11-25
[FIX] BroadcastWithAck behavior

BCCSTeam:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
On remote sockets don't call callback with arguments.
Returns null as error and data.

### New behavior
On remote sockets with call callback with arguments.
Now it returns error and data from callback.


lib/broadcast-operator.ts
```diff
@@ -252,10 +252,8 @@ export class BroadcastOperator<EmitEvents extends EventsMap, SocketData>
         responses.length === expectedClientCount
       ) {
         clearTimeout(timer);
-        ack.apply(this, [
-          null,
-          this.flags.expectSingleResponse ? null : responses,
-        ]);
+        if (this.flags.expectSingleResponse) ack.apply(this, responses);
```
darrachequesne, line 255: Shouldn't it be `if (this.flags.expectSingleResponse) ack.apply(this, responses[0]);` in this case?

So that:

```js
const [remoteSocket] = await io.in(theSocketId).fetchSockets();

remoteSocket.timeout(1000).emit("hello", (err, response) => {

});

```

Behaves like:

```js
  socket.timeout(1000).emit("hello", (err, response) => {

  });
});
```

What do you think?

Issue #4882, opened 2023-11-28
Unexpected behavior from catch-all listeners.


NyaStone:
I've been working on some middleware for data validation and infering the event types using a zod schema.
When looking for how to handle acknowledgements through my middleware I found a few unexpected behavors of catch-all listener.

It is possible to add some logic before acknowledging a client's request.
```ts
socket.use(packet: [string, ...any[]], next:(err?: Error) => void) {
    console.log(packet);
    // [ 'eventWithAck', [Function (anonymous)] ]
    next();
};

socket.on('eventWithAck', (ack: () => void) => {
    ack();
});

client.emit('eventWithAck', () => {
    console.log('done');
});
```
Here we can get access to the callback that is responding to the client, and we can wrap it with our own logic to be run right before the acknowledgement is sent.
```ts
socket.use(packet: [string, ...any[]], next:(err?: Error) => void) {
    if (packet.length && typeof packet[packet.length-1] === 'function') {
        const callback: (...args: any[]) => void = packet[packet.length-1];
        packet[packet.length-1] = (...args: any[]) => {
            console.log(args);
            // [ 'some data' ]
            callback(...args);
        }
    }
    next();
};

socket.on('eventWithAck', (ack: (data: string) => void) => {
    ack('some data');
});

client.emit('eventWithAck', () => {
    console.log('done');
});
``` 

The issue is that this is only achievable to check acknledgement that we send out, not the incomming acknowlegements data packets. 
I had expected to achieve a similar logic using the catch-all listeners.
The documentation on the Socket.io website stated that catch all listeners don't trigger upon recieving/sending acknoledgements, but i hoped being able to get a reference to the acknowledgement callback before the emit is being sent, so i could wrap it with my logic. Wrapping the acknowledgement callback there allows me to also have context of the initial emited event (before it's sent) for tha added logic.
However when I tried i quickly noticed that the acknowledgement callback gets stripped away from the data packet and sent to the acknowledgement listeners before the catch-all listeners get called.
Second issue with this proposition is that I don't know if the spread arguments would pass the callback by reference so it can be wrapped.

While writing this I noticed that the documentation in VS code of catch all listeners does warn me about the callback being stripped.

The solution proposed here https://github.com/socketio/socket.io/issues/4649 seems like migration hell because of potential collisions when suddenly triggering the listeners for new reasons.

[...]

PR #4885, opened 2023-11-30
Adding support for middlewares intercepting outgoing events

NyaStone:

* [ ] a bug fix
* [x] a new feature
* [ ] other

### Current behavior

Currently middlewares are limited to hook before the server aknowledges a clients message, but it can't hook on the client acknowledging from the client.
Currently data validation logic needs to be rewritten in the logic of every emited with acknowledgements, that may result in a lot of boiler plate code.

### New behavior

Introduced the useOutgoing((event: [string, ...any[]], next: (err?: Error) => void) => void) method.
It is behaving much like regular use() does, but before the messages are being sent from the server to the client.
(They do run before the catch all listeners and before the acknowledgement callback gets popped to the ack manager in order to be able to edit the event, or abort it).

Example: 
```ts
    socket.useOutgoing((event, next) => {
      if (event.length && typeof event[event.length - 1] === 'function') {
        const callback = event[event.length - 1];
        event[event.length - 1] = (...args) => {
          // able to hook right before the acknowledgement callback is executed.
          callback(...args);
        }
      }
      // do not forget to call next
      next();
    });
});
```

I disscussed other option briefly in this issue https://github.com/socketio/socket.io/issues/4882.
For short, the tricky part about this, is that one might want access to the context of the event that originated the ack repsonse whilst consuming what the client acknowledged. Else it would be of no use hooking there, except for logging purposes I guess.
This would be a solution that doesn't disrupt current usage of Socket.IO, and we have access to the context of the event.

### DIFF
--- a/lib/socket.ts
+++ b/lib/socket.ts
@@ -236,6 +236,8 @@ export class Socket<
   private readonly adapter: Adapter;
   private acks: Map<number, () => void> = new Map();
   private fns: Array<(event: Event, next: (err?: Error) => void) => void> = [];
+  private outFns: Array<(event: Event, next: (err?: Error) => void) => void> =
+    [];
   private flags: BroadcastFlags = {};
   private _anyListeners?: Array<(...args: any[]) => void>;
   private _anyOutgoingListeners?: Array<(...args: any[]) => void>;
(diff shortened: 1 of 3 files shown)


bperel:
Is this PR basically allowing to add event emit interceptors? It could be a great way to allow acknowledgement caching, in the same way that [axios-cache-interceptor](https://github.com/arthurfiorette/axios-cache-interceptor) does it for Axios requests :-)

NyaStone:
> Is this PR basically allowing to add event emit interceptors? It could be a great way to allow acknowledgement caching, in the same way that [axios-cache-interceptor](https://github.com/arthurfiorette/axios-cache-interceptor) does it for Axios requests :-)

I haven't played with axios around yet. But from what I read from the [axios-cache-interceptor](https://github.com/arthurfiorette/axios-cache-interceptor) docs the idea of this PR is similar to the interceptors, even tho the implementation varies a little bit. I based my implementation on the socket.io socket middlewares implementation so it works the same  (unlike axios cache interceptors where inbound interceptors are FILO and outblound are FIFO).
The main point stands of being able to hook into and filter properly in and outbound requests is the same.

PR #4907, opened 2023-12-26
Change 'import express = require("express");' to correct TS syntax

mivpt:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

import express = require("express")
### New behavior

import express, { type Request } from "express";


darrachequesne:
I don't think this is right. Don't you need to set `allowSyntheticDefaultImports` to `true` in that case?

darrachequesne:

Issue #4914, opened 2024-01-05
Calling io.emit("event") with events that has no arguments result in TypeScript errors


lts20050703:
**Describe the bug**
Calling io.emit("event") with events that either no argument result in TypeScript errors.

**To Reproduce**

Socket.IO server version: `4.7.3`

*Server*

```ts
import { Server } from "socket.io";

interface ServerToClientEvents {
  noArg: () => void;
}

const io = new Server<{}, ServerToClientEvents>();

io.emit("noArg"); // <- Error
```

**Expected behavior**
No errors

**Actual behavior**
`Argument of type 'string' is not assignable to parameter of type 'never'. ts(2345)`

**Platform:**
 - Device: Lenovo IdeaPad 5 14ITL05
 - OS: Windows 11 Home 23H2 (22631.2861)

**Additional context**
Seems to be a regression, after downgrading to 4.7.2 the error disappeared.

lts20050703:
This seems to be caused by f6ef267b035a4db49b7d0fce438ca5c5b686f547

ZachHaber:
@lts20050703 - from https://github.com/socketio/socket.io/issues/4813#issuecomment-1717492560

> I think most of the issues come from the fact that calling .timeout() is (at least currently) mandatory on the server side, since there is no default timeout value (the client has an ackTimeout option though).
> 
> So, for example:
> ```js
> // INVALID (will time out directly anyway)
> io.emit('withAck', 'hello', (...args) => { });
> // VALID
> io.timeout(1).emit('withAck', 'hello', (...args) => { });
> ```

For emitting with ack from the server, it is correctly an error now, because it wouldn't do what you'd expect from the beginning and would timeout immediately as [darrachequesne](https://github.com/darrachequesne) pointed out. Plus the type definition was misleading for it in 4.7.2 anyway. `e` would actually be a `number[]` rather than a `number` - if it was valid at all.

> `io.emit("withAck", "4", (e) => {}); // <- Error`

I will however look into the no-callback one, since that should be valid.

ZachHaber:
@lts20050703, PR is up with the fix for the noArgs issue.

Until the fix is merged in and a new version is cut, you can use [patch-package](https://www.npmjs.com/package/patch-package) (or something similar) to update the `dist/typed-events.d.ts` file.


```diff
[...]

PR #4915, opened 2024-01-05
Fix: Calling io.emit("event") with events that have no arguments results in TypeScript errors

ZachHaber:

* [x] a bug fix
* [ ] a new feature
* [ ] other

Typescript type fixe relating to #4817, also type failures in the tests due to the client#id type becoming `string|undefined` ([f9c16f2](https://github.com/socketio/socket.io-client/commit/f9c16f226512fc8a8df461e3a07e392720462165))

### Current behavior
```ts

interface ServerToClientEvents {
  noArg: () => void;
}

const io = new Server<{}, ServerToClientEvents>();

io.emit("noArg"); // <- Error
```

### New behavior
```ts

interface ServerToClientEvents {
  noArg: () => void;
}

const io = new Server<{}, ServerToClientEvents>();

io.emit("noArg"); // ✔️ - no error!
```

#4914

### DIFF
--- a/lib/typed-events.ts
+++ b/lib/typed-events.ts
@@ -52,9 +52,11 @@ export type EventNamesWithoutAck<
 > = IfAny<
   Last<Parameters<Map[K]>> | Map[K],
   K,
-  K extends (
-    Last<Parameters<Map[K]>> extends (...args: any[]) => any ? never : K
-  )
+  K extends (Parameters<Map[K]> extends never[] ? K : never)
+    ? K
+    : K extends (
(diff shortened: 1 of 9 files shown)


lts20050703:
Hmm does this also fix #4917?

ZachHaber:
Yes, they all use the same utility type to set up which events are allowed. I just added in a test I was missing for a namespace emit with noArgs.

darrachequesne:
@ZachHaber thanks :+1:

PR #4931, opened 2024-01-23
Add emitWithAck method to RemoteSocket class

verycosy:

* [ ] a bug fix
* [x] a new feature
* [ ] other

### Current behavior
`fetchSockets()` returns `RemoteSocket` array.
But `RemoteSocket` does not have `emitWithAck` method.
So currently, we can not use handy way.

### New behavior
`RemoteSocket` instances can use `emitWithAck` method also.

I just added code like `emit` method.
It's calling `BroadcastOperator`'s `emit` method.

### DIFF
--- a/lib/broadcast-operator.ts
+++ b/lib/broadcast-operator.ts
@@ -529,6 +529,13 @@ export class RemoteSocket<EmitEvents extends EventsMap, SocketData>
     return this.operator.emit(ev, ...args);
   }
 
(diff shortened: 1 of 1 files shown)


verycosy:
it would be better just using `timeout`.
when `timeout` method is called, `RemoteSocket` can call `emitWithAck`

ex) `socket.timeout(1000).emitWithAck('event')`

PR #4950, opened 2024-02-18
parent-namespace: change inline Adapter-like to file-private class

jokester:

* [ ] a bug fix
* [ ] a new feature
* [x] other: remove 1 FIXME in code

### Current behavior

`parentNamespace.adapter` be a dummy adapter-like object, and only supports `broadcast` operation.

### New behavior

The same behavior, but with a class that will pass type check.

When reading the code I saw the FIXME, so I thought the type check can be satisfied this way.

IMO except the type check the inline `{broadcast}` adapter was not bad. It is arguably better in that unsupported operations will fail loudly instead of silently. So I'm not very confident of this PR. Opinions welcome 🙏🏽 .

To pass type check while preserving the old fail-fast behavior, one could also `this.adapter = {broadcast} as Adapter;`

### DIFF
--- a/lib/parent-namespace.ts
+++ b/lib/parent-namespace.ts
@@ -6,6 +6,7 @@ import type {
   DefaultEventsMap,
   EventNamesWithoutAck,
 } from "./typed-events";
+import {Adapter} from "socket.io-adapter";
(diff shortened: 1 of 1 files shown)


darrachequesne:
@jokester great, thanks :+1:

PR #4971, opened 2024-03-11
correctly await async close on adapters

marknelissen:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
The close of the server does not await the async close operations of the adapters like the MongoAdapter, leading to errors due to premature closing of the underlying connection.

### New behavior
The close awaits the async operations on the adapters of the different namespaces. This allows to be sure that e.g. the Mongo connection can be safely disposed of.

I've chose the `Promise.allSettled` to avoid being too disruptive, since the current synchronous implementation already disregards any errors the closing of the adapters might raise.


darrachequesne:

In that case, do you think it would make sense to wait for the completion of `this.httpServer.close()` below too?

Reference: https://nodejs.org/api/http.html#serverclosecallback

marknelissen:
That is already what is done. The only way to wait for it, is to pass the callback function, which the current code already does. It is not an awaitable function, since it does not return a Promise.

BrandonZacharie:
bump

BrandonZacharie:
@darrachequesne What else is needed to get this merged?

marknelissen:
@darrachequesne I resynced the branch. Can you approve?

darrachequesne:
@marknelissen thanks :+1:

Issue #3507, opened 2019-11-02
Prevent socket io from emitting ping timeout for inactive or not focused browser page in mobile browsers chrome, firefox and safari.


FaizanZahid:
### You want to:

* [x] report a *bug*

### Current behaviour

*What is actually happening?*

I have a simple chat app using socket io. Very similar to the socket io chat official demo. Socket io goes to ping timeout on all mobile browsers whenever the user minimizes the page or opens another app or page gets out of foucs.

In my web app whenever, the user clicks on file input upload from their chrome/firefox/safari browser on mobile, it opens up the phone dialog box to select the file from galley to select and upload. During this time (my chat page gets out of focus due to visibility), when user is searching for a file, and after exact 30 seconds, socket io client issues ping timeout, and the person chat is disconnected.

In my application level, i am doing a reconnect if socket io goes to ping timeout for whatever reason, but that doesn't fix my problem, because if the user clicks on 1 file after 31 seconds, then that image file is not sent in chat, because socket io was timed out during that time, and it takes few seconds to reconnect, when the user's focus is back on the browser tab. So the file upload action is lost, because file is uploaded but socket io connection was on ping timeout. so when it reconnects, the selected file upload is never sent to the other connected user.
### Steps to reproduce (if the current behaviour is a bug)

Please check relevant stackoverflow question: 
https://stackoverflow.com/questions/58667207/prevent-socket-io-from-ping-timeout-on-inactive-or-not-focused-browser-page-on-c

### Expected behaviour

Whenever on mobile or desktop browsers, if a user navigates away from socket io chat page and comes back, socket io should still keep the client connected, and should not issue a ping timeout event and cause disconnect after 30 seconds. It should keep chat active no matter how long a person is out of focus of the page.

*What is expected?*

### Setup
- OS: Android and iOS
- browser: happening on safari, firefox, chrome (android and iphones)
- socket.io version: 2.2.0 (same issue happening on 2.3.0 too, i tried upgrading my client and socket io both)

### Other information (e.g. stacktraces, related issues, suggestions how to fix)

I tried detecting whenever user's page focus is out, using this code,

document.addEventListener("visibilitychange", onchange);
from Is there a way to detect if a browser window is not currently active?

[...]

PR #5134, opened 2024-07-15
fix: Check if missed heartbeat synchronously before sending message

tjenkinson:
Chrome (and maybe other browsers) will throttle `setTimeout` under certain conditions, causing them to slow down, and this breaks the heartbeat detection logic.

This fix means if a message is emitted when the connection should be considered expired (but the `setTimeout` hasn't fired yet), we close that socket and then buffer the message for the new connection. So now a message should never be sent over an expired connection.

This doesn't solve the problem of connections being detected as expired too late. I think that could be done separately.

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
Check the repro on https://github.com/socketio/socket.io/issues/5135

### New behavior
- When the next message is sent, the system will realise it's missed the heartbeat, create a new connection, and then the message will be sent over the new connection
- So now a message will never be sent over a stale connection, and we're not depending on the heartbeat timeout to fire correctly when a page comes back to life (which doesn't happen when it's throttled)

The reason that the heartbeat logic is not immediately closing the old connection when the page is woken up is due to chrome pausing the `setTimeout` that's used in the heartbeat logic. This PR doesn't fix that. I Think that could be handled separately.

https://github.com/socketio/socket.io/issues/3507#issuecomment-977681297 is another fix related to throttled timers

Let me know if you think this makes sense and I'm happy to look into adding a test for it


packages/socket.io-client/lib/socket.ts
```diff
@@ -440,6 +440,9 @@ export class Socket<
       packet.id = id;
     }
 
+    // check synchronously if we've missed a heartbeat, potentially causing a disconnection
+    this.io.checkHeartbeat()
```
darrachequesne, line 444: How about merging the `checkHeartbeat()` method with the `isTransportWritable` check below? Something like:

`packages/engine.io-client/lib/socket.ts`

```js
get writable() {
  const expired = Date.now() >= this._pingTimeoutTime
  if (expired) {
    debug("`checkHeartbeat` detected missed ping so closing connection");
    this._onClose("ping timeout");
    return false;
  }

  return this.io.engine.transport?.writable;
}
```

`packages/socket.io-client/lib/socket.ts`

```js
const isTransportWritable = this.io.engine.writable;
```

tjenkinson: It feels a bit strange that reading `writable` can have a side effect? Especially when it can cause events to be emitted as part of that

Maybe `writable` should return `false` when a heartbeat has been missed, but we should still have a function like `checkHeartbeat` (or maybe `disconnectIfHeartbeatMissed()`) that does the disconnection.

WDYT?

darrachequesne: With the `checkHeartbeat()` method, it feels like we are mixing two different layers. In my opinion, the Socket.IO client should not have to be aware of the internals of the Engine.IO client (up to a certain point, obviously).

That's why I suggested to merge the heartbeat check with the transport check below. Maybe not as a getter with side effects, but how about a method like `socket.isReady()`?

tjenkinson: I see. I liked the `writable` idea but I think it was the synchronous disconnection as part of reading that that felt a bit strange.

How about:
- adding `isAlive`/`isResponsive` getter which will check the last heartbeat timestamp, and if it’s `false` also schedule a disconnect on the next tick
- Maybe update `writable` too to also check the last heartbeat time and schedule a disconnection
- Instead of `this.connected` check the `isAlive` instead. (`this.connected` doesn’t work if the disconnect gets triggered on the next tick)

I think it feels better doing the disconnection on the next tick instead of sync as part of the read call, as then it’s the same pattern as if the heartbeat timer task fired anyway and there can’t be event listeners being called as part of the read

WDYT?

---

Or

We don't add `isResponsive`, and make `writable` do the check and queue the disconnect, but update the `.connected` check below to instead check `.writable`, so we always buffer if it's not writable. Not sure what impact this might have?

darrachequesne: Sorry for the delay.

> adding isAlive/isResponsive getter which will check the last heartbeat timestamp, and if it’s false also schedule a disconnect on the next tick

That sounds reasonable :+1:

Note: we already have a `nextTick()` method [here](https://github.com/socketio/socket.io/blob/582655f679ccc43f0a9cbef1f13ea3cde07dc2e1/packages/engine.io-client/lib/globals.ts#L1-L9).

> We don't add isResponsive, and make writable do the check and queue the disconnect, but update the .connected check below to instead check .writable, so we always buffer if it's not writable. Not sure what impact this might have?

Not sure about this one, since `writable` is an attribute of the `Transport` class, while checking the heartbeat expiry sounds like the duty of the `Socket` class.

tjenkinson: ok great thanks will update to that first approach then

tjenkinson: @darrachequesne I've pushed the changes for this

packages/engine.io-client/lib/socket.ts
```diff
@@ -708,6 +715,32 @@ export class SocketWithoutUpgrade extends Emitter<
...
   }
 
+  /**
+   * Returns `true` if the connection is responding to heartbeats.
+   *
+   * If heartbeats are disabled this will always return `true`.
+   *
+   * @return {boolean}
+   */
+  public isResponsive() {
```
darrachequesne, line 725: To be honest, I'm not sure about this public method. This is something we'll have to maintain forever, and which is not part of the WebSocket API which we try to mimic (more or less). I'm not sure how we could implement it differently though...

packages/engine.io-client/lib/socket.ts
```diff
@@ -717,6 +750,9 @@ export class SocketWithoutUpgrade extends Emitter<
    * @return {Socket} for chaining.
    */
   public write(msg: RawData, options?: WriteOptions, fn?: () => void) {
+    // this will schedule a disconnection on next tick if heartbeat missed
+    this.isResponsive();
```
darrachequesne, line 754: Is this really needed? The write will fail anyway, no?

packages/socket.io-client/lib/socket.ts
```diff
@@ -449,7 +449,10 @@ export class Socket<
       this.flags.volatile && (!isTransportWritable || !this.connected);
     if (discardPacket) {
       debug("discard packet as the transport is not currently writable");
-    } else if (this.connected) {
+    } else if (
+      this.connected &&
+      (this.flags.volatile || this.io.isResponsive())
```
darrachequesne, line 454: Is the `volatile` check needed here?

packages/engine.io-client/lib/socket.ts
```diff
@@ -12,6 +12,7 @@ import {
   defaultBinaryType,
 } from "./globals.node.js";
 import debugModule from "debug"; // debug()
+import { nextTick } from "./globals.js";
```
darrachequesne, line 15: You can import the `nextTick` method from "./globals.node.js", so that it works in all environments.

packages/engine.io-client/lib/socket.ts
tjenkinson: The problem is without this the write will fail but then the packet is lost and won’t be retried

With this the packet gets queued and sent when the connection is restored

packages/socket.io-client/lib/socket.ts
tjenkinson: I thought it made sense because doesn’t `volatile` mean you either want the packet sent immediately or discarded? So we could try and send it regardless if whether a heart beat has been missed because if it fails it doesn’t matter

packages/engine.io-client/lib/socket.ts
tjenkinson: Looks like things like `_packet` are fake private? Could prefix it with an underscore and document it to be an internal api?

another potential idea was to add a return value to `_packet` which signals if the packet was actually written

packages/engine.io-client/lib/socket.ts
tjenkinson: won't that make it only work in node? That file is hard coded to `process.nextTick`

https://github.com/socketio/socket.io/blob/fd99f2e15fef0e17ae9e3849c92ef0ca616a5919/packages/engine.io-client/lib/globals.node.ts#L1

tjenkinson: ah right

darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/8adcfbfde50679095ec2abe376650cf2b6814325.

I've made some minor edits and added a few tests. Could you please check if everything makes sense?

Issue #5139, opened 2024-07-18
Emit to room not work with uWebsocketjs


SaltFish001:
**Describe the bug**
Emit some message to room not work with uWebsocketjs, but clients does`t got any message. 
**To Reproduce**

Please fill the following code example:

Socket.IO server version: `x.y.z`

*Server*

```js
import { Server as socket_server } from 'socket.io';
import { App } from 'uWebSockets.js';

const uapp = App();
const socket_instance = new socket_server({
  path: '/socket',
});
socket_instance.attachApp(uapp);
socket_instance.use((scoket, next) => {
  console.log('socket connect >>>', scoket.id);
  scoket.join('test');
  setTimeout(() => {
    console.log('socket emit >>>', scoket.rooms);
    socket_instance.to('test').emit('message', 'hello world');
  }, 2000);
  return next();
});

uapp.listen(9999, (token) => {
  console.log(token);
  console.log('server is listening on port 8888');
});

```

Socket.IO client version: `x.y.z`

*Client*

```js
import { io } from 'socket.io-client';
const client = io('http://localhost:9999', {
  path: '/socket',
  transports: ['websocket'],
});

client.onAny((...arg) => {
  console.log('onany >>>', arg);
});
client.connect();

```

**Expected behavior**

**Platform:**
 - Device: [e.g. Samsung S8]
 - OS: [e.g. Android 9.2]

**Additional context**
Socket.io & Socket.io-client 4.7.5
uWebsocketjs 20.44.0

darrachequesne:
Hi! I was indeed able to reproduce the issue.

It does not seem to happen with `uWebSockets.js#v20.30.0` though, this needs some additional investigation.

SaltFish001:
> Hi! I was indeed able to reproduce the issue.嗨！我确实能够重现这个问题。
> 
> It does not seem to happen with `uWebSockets.js#v20.30.0` though, this needs some additional investigation.然而，这种情况在 身上似乎并未出现，这需要进一步的调查。

Thanks reply. Right now I can use fetchSockets get all socket , then foreach  sockets to emit event. Hope this is not a hard work for you.

TXWSLYF:
I found that if you add one line of code：
```js
const { Server: socket_server } = require("socket.io");
const { App } = require("uWebSockets.js");

const uapp = App();
const socket_instance = new socket_server({
  path: "/socket",
});
socket_instance.attachApp(uapp);
socket_instance.use((scoket, next) => {
[...]

PR #5158, opened 2024-08-07
fix: track socket before run middlewares

TXWSLYF:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
the socket is not tracked when run the middlewares

### New behavior
track socket before run middlewares

fix: https://github.com/socketio/socket.io/issues/5139

### DIFF
--- a/packages/socket.io/lib/namespace.ts
+++ b/packages/socket.io/lib/namespace.ts
@@ -322,6 +322,9 @@ export class Namespace<
     debug("adding socket to nsp %s", this.name);
     const socket = await this._createSocket(client, auth);
 
+    // track socket
+    this.sockets.set(socket.id, socket);
+
     if (
       // @ts-ignore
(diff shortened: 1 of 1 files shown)


packages/socket.io/lib/namespace.ts
```diff
@@ -322,6 +322,9 @@ export class Namespace<
     debug("adding socket to nsp %s", this.name);
     const socket = await this._createSocket(client, auth);
 
+    // track socket
+    this.sockets.set(socket.id, socket);
```
darrachequesne, line 326: Hmm, I'm not sure about this change, since the `sockets` map currently contains only sockets that are actually connected (not in the process of connection).

darrachequesne:
Superseded by https://github.com/socketio/socket.io/commit/b04fa64365729244a9c50a6b54b12e9bcc9e55d0. Thanks :+1:

PR #5169, opened 2024-08-12
remove prettier v2

jokester:

* [ ] a bug fix
* [ ] a new feature
* [x] other

### Current behavior

2 versions of prettier, respectively v2 and v3, get installed.

### New behavior

Only use prettier v3

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,0 +1,3 @@
+{
+  "trailingComma": "es5"
+}
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,3 +1,0 @@
-{
-  "trailingComma": "es5"
-}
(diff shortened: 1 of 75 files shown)


package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "name": "socket.io",
```
jokester, line 2: This change makes `package-lock.json` to contain a stable root package name (otherwise it uses the directory name)

.prettierrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "trailingComma": "es5"
```
jokester, line 2: I don't have strong preference on prettier config. It's just this options creates fewer diffs.

darrachequesne:
@jokester thanks a lot :+1: I've removed the custom configuration, because I think the main purpose of prettier is to be opinionated.

Issue #5187, opened 2024-09-12
Specify transport type or autocompletion.


nayounsang:
**Is your feature request related to a problem? Please describe.**
```
import { io } from "socket.io-client";
    const newSocket = io(process.env.HELLO, {
      withCredentials: true,
      transports: [], // something...
    });
```

**Describe the solution you'd like**
I was able to specify specific transports.
I think it would be more convenient if it was auto-completed.

**Describe alternatives you've considered**
A different type instead of string[]
**Additional context**

darrachequesne:
For future readers:

This should be fixed by https://github.com/socketio/socket.io/commit/b5ccfd4838b901c7c4efd1e5e07a0bd083620619, included in `engine.io-client@6.6.1` and `socket.io-client@4.8.0`.

PR #5188, opened 2024-09-12
refactor: default transport in type

nayounsang:

* [ ] a bug fix
* [ ] a new feature
* [x] other

### Current behavior
ah! I tried to specify transports, but it doesn't autocomplete.
Because current type of transport is string[].
I have to check JSDOC to see  ['polling','websocket', 'webtransport'].

### New behavior
Be comfortable with auto-completion without typos.

#5187


darrachequesne:
@nayounsang thanks!

Issue #5179, opened 2024-08-30
Namespaces are contravariant over SocketData because `Namespace['_fns']` is not `private`


ammut:
**Describe the bug**

I have a package that exports a function requiring a typed Namespace to do its thing. This Namespace has a `SocketData` value of `{ connectionId: string }`. In my actual application I then set up such a namespace to provide said setup function with a such a namespace.

However, because the actual namespace I set up has additional properties in `SocketData`, for other parts of the application. This results in a TypeScript compiler error when trying to apply said actual namespace to that setup function.

The error stems from the fact that the `_fns` property on `Namespace` is [contravariant](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#optional-variance-annotations-for-type-parameters) over `SocketData`.

**To Reproduce**

*Server*

```ts
import type { Namespace } from "socket.io";
import { Server } from "socket.io";

const io = new Server(3000, {});
const ns = io.of('/my-ns') as Namespace<any, any, any, { connectionId: string; otherStuff: string }>;

const nsForMyPackage: Namespace<any, any, any, { connectionId: string }> = ns; // <- THIS ERRORS

```

**Additional context**

*In theory*, because `_fns` is not marked `private`, I can do something like this:

```ts
declare const nsA: Namespace<any, any, any, { foo: string; bar: string }>;

nsA.use((socket, next) => {
  console.log(socket.data.bar.toUpperCase());
  next();
});

const nsB: Namespace<any, any, any, { foo: string }> = nsA; // BAD ASSIGNMENT

const nsC: Namespace<any, any, any, { foo: string }>;

nsC.on('connection', (socketC) => {
  nsB._fns[0](socketC, () => {}); // BAD CALL
});
```

To the TypeScript compiler, `BAD CALL` looks fine. `nsB._fns` contains functions that require a Socket with `SocketData` of `{ foo: string }`. `nsC.on('connection', ...)` provides just such a Socket. However, that middleware will try to call `socket.data.bar.toUpperCase()` and produce a `TypeError: bar is undefined`. That is why the compiler warns me at `BAD ASSIGNMENT`.

HOWEVER, I should never actually know anything about `_fns`, even less access it. That would make it impossible to do any such mischief. Then the assignment at `BAD ASSIGNMENT` would be safe. As even in that case, code like this would be sound:

```ts

declare const nsA: Namespace<any, any, any, { foo: string; bar: string }>;
nsA.use((socket, next) => {
  // setup socket data
[...]

PR #5196, opened 2024-09-24
fix: make Namespace._fns private

ammut:

* [x] a bug fix
* [ ] a new feature
* [ ] other

Issue: #5179

**This is a variant for PR #5195, please only merge one of these!** I created two PRs for the related issue, providing both suggested solutions. This one uses the index signature trick as suggested in the issue.

I tried adding a test to prevent regression, but didn't manage to reproduce the error that caused all this inside the socket.io test suite. I assume this is because we use a stricter TypeScript configuration in our own codebase, but didn't bother verifying it.

### DIFF
--- a/packages/socket.io/lib/namespace.ts
+++ b/packages/socket.io/lib/namespace.ts
@@ -161,7 +161,7 @@ export class Namespace<
     SocketData
   >;
 
-  protected _fns: Array<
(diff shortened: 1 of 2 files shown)


darrachequesne:
@ammut thanks :+1:

PR #5198, opened 2024-09-25
comment: write js docs in on

nayounsang:

* [ ] a bug fix
* [x] a new feature
* [x] other

### Current behavior
- About `socket-io-client-emitter/Emitter` `on` method.
- It was difficult for us to know what the reserved event names were without looking at the docs.

### New behavior
image
- I thought it would be better to use JSDoc without changing the types. This is because there is an advantage of auto-completion when using union, etc., but developers often use custom event names, so I thought the type could actually cause confusion.
- So I used JSDoc. If the user checks the JSDoc, they can see the scheduled event name and its brief description.
If developers are curious about more, I have linked [the docs](https://socket.io/docs/v4/client-api/#events-1).

- If you want to edit, change or delete content, please always leave a comment.
- I've only written about `on` at the moment, so if you think it would be good to write about something else as well, please let me know.

### DIFF
--- a/packages/socket.io-client/lib/socket.ts
+++ b/packages/socket.io-client/lib/socket.ts
@@ -120,7 +120,7 @@ interface SocketReservedEvents {
   connect_error: (err: Error) => void;
   disconnect: (
     reason: Socket.DisconnectReason,
-    description?: DisconnectDescription
+    description?: DisconnectDescription,
   ) => void;
 }
 
@@ -555,7 +555,7 @@ export class Socket<
           debug(
             "packet [%d] is discarded after %d tries",
             packet.id,
-            packet.tryCount
+            packet.tryCount,
           );
           this._queue.shift();
           if (ack) {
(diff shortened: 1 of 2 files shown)


darrachequesne:
Hi! Thanks a lot for your PR.

I'm not sure this is right though, as the emitter provided by the `@socket.io/component-emitter` is used in several places, not just for the `Socket` object of the `socket.io-client` package.

Another possibility would be to override the `on` method in the `Socket` class, and add the improved documentation there. What do you think?

nayounsang:
Hi. I commit this. 
Is this correct what you said?
- I was trying to use `Prettier` just for my work, but at some point it started applying to all my files. A few `,` are missing.
- I was confused about where to place the overridden method. First of all, since it is a new change, I placed it at the very bottom.
- I am wondering how to use the `ReversedEvents` type defined as a generic in the `Emitter`. Basically, it simply inherits EventMaps, so I created a new interface.

Issue #5258, opened 2024-12-12
Client emits initial event twice inside 'connect' handler when retries are enabled.


backendsuraj:
**Describe the bug**

When the retries option is enabled in the Socket.IO client, the client emits the initial event twice inside the 'connect' handler. However, this issue does not occur when the retries option is not provided.

**To Reproduce**

Please fill the following code example:

Socket.IO server version: `4.8.1`

*Server*

```js
const io = new Server(httpServer, {});

io.on("connection", (socket) => {
  console.log(`connect ${socket.id}`);

  socket.on('data', (data, callback)=> {
    console.log('Incoming message from client: ', data);
    callback({
      success: true,
      message: 'Acknowledged...'
    });
  });

  socket.on("disconnect", (reason) => {
    console.log(`disconnect ${socket.id} due to ${reason}`);
  });
});
```

Socket.IO client version: `4.8.1`

*Client*

```js
const socket = io(`http://localhost:${port}`, {
  retries: 2,
  ackTimeout: 3000,
  transports: ["websocket"],
});

socket.on("connect", () => {
  console.log(`connect ${socket.id}`);
  const msg1 = 'Hello!!!';
  const msg2 = 'Byeee';
  socket.emit('data', msg1);
  socket.emit('data', msg2);
  console.log('Emitted msg: ', msg1);
  console.log('Emitted msg: ', msg2);
});
```

**Expected behavior**
The event inside the 'connect' handler should be emitted only once, regardless of whether the retries option is enabled or disabled.

**Platform:**
 - Device: Windows/google chrome
 - OS: Windows/google chrome

**Additional context**
Console output:
![image](https://github.com/user-attachments/assets/483e449d-4011-41f9-a43c-5b1af6255fbf)

I debugged this issue and it seems like:

After the client emits the connect event, the event queue is drained with the force option set to true. This action retriggers events that were already added to the queue & awaiting acknowledgement due to the retries option. As a result, the initial event is emitted twice.

![image](https://github.com/user-attachments/assets/c4274a2f-bee4-415c-9512-d614244bd7d1)

darrachequesne:
Hi! I think that's because the client expects a `null` as a first argument for the callback:

```js
io.on("connection", (socket) => {
  // [...]
  socket.on('data', (data, callback)=> {
    console.log('Incoming message from client: ', data);
    callback(null, {
      success: true,
      message: 'Acknowledged...'
    });
  });
  // [...]
});
```

[...]

PR #5259, opened 2024-12-12
Drain queue before emitting connect event.

backendsuraj:

* [x] a bug fix
* [ ] a new feature
* [ ] other

Fixes: #5258 

### Current behavior
When the retries option is enabled in the Socket.IO client, the client emits the initial event twice inside the 'connect' handler.

### New behavior
The event inside the 'connect' handler should be emitted only once, regardless of whether the retries option is enabled or  disabled.

We can drain the queue before emitting the connect event.


darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/d19928e8d8b325310274031ed7de2ddc93ebb589. Sorry for the (very) late merge...

Issue #5340, opened 2025-04-28
It it duplicating the self messages after refreshing the window in examples/private messages


MeeeeetShah:
**Describe the bug**
It it duplicating the self messages after refreshing the window in examples/private messages 

**To Reproduce**

Go through examples and take pull from 4th (final branch) for example/private message if we send message to self then it seems to be correct but after refreshing the window we see duplicating the self messages.. 

Socket.IO server version: `3.1.2`

*Server*

```js
import { Server } from "socket.io";

const io = new Server(3000, {});

io.on("connection", (socket) => {
  console.log(`connect ${socket.id}`);

  socket.on("disconnect", () => {
    console.log(`disconnect ${socket.id}`);
  });
});
```

Socket.IO client version: `4.8.1`

*Client*

```js
import { io } from "socket.io-client";

const socket = io("ws://localhost:3000/", {});

socket.on("connect", () => {
  console.log(`connect ${socket.id}`);
});

socket.on("disconnect", () => {
  console.log("disconnect");
});
```

**Expected behavior**
It should show single message only

darrachequesne:
Fixed by https://github.com/socketio/socket.io/commit/5a60fd2b7cedd9ad0511d91c751d9f8618b92d91.

PR #5341, opened 2025-04-29
Update messageStore.js to fix Issue:It it duplicating the self messag…

MeeeeetShah:
…es after refreshing the window in examples/private messages #5340

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

It's duplicate the message in redis store for self messages due to no condition checks
### New behavior

It resolved by just adding a simple condition


darrachequesne:
Awesome, thanks! (and sorry for the delay)

Issue #4546, opened 2022-12-07
Socket server tries to close external http server, causing `ERR_SERVER_NOT_RUNNING` error


panta82:
**Describe the bug**

When you initialize socket server, you can either give it an external instance of `HTTPServer`, or let socket.io create its own internal instance.

However, in either case, SocketIO tries to close the server when shutting down:

https://github.com/socketio/socket.io/blob/3b7ced7af7e0a2a66392577f94af1ee5ed190ab1/lib/index.ts#L685

In my case, I have my own shutdown procedure elsewhere, and it causes server to throw `ERR_SERVER_NOT_RUNNING` error, because server.close() was called twice.

This might be appearing now due to new stricter behavior in node.js, I am not sure. This is running on `v14.17.3`.

**Expected behavior**

IMO socket.io should only close HTTP server if it's its own instance. If an instance is provided, it's someone else's responsibility to close that server. Socket.IO shouldn't interfere.

**Mitigation**

I am currently using this hack to prevent the uncaught exception:

```
ioServer['httpServer'] = null;
ioServer.close();
```

PR #5344, opened 2025-04-30
fix(socket.io): improve `close()` function

fantua:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

After https://github.com/socketio/socket.io/pull/4971 the usage of the `close()` function has become confusing and sometimes misleading.

Since `close()` now returns a Promise, I would expect the `httpServer` to close when the promise is resolved. However, this is not the case.

```ts
await io.close();
// httpServer not closed yet
```

Also, in order to catch all possible errors, I need to use a mixed style of async/await and callbacks:

```ts
io.close(error => {
  if (error) {
    // error here
  }
}).catch(error => {
  // another error here
});
```

### New behavior

```ts
await io.close();
// httpServer closed
```

```ts
io.close().catch(error => {
  // all errors here
});
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15507,7 +15507,7 @@
       }
     },
     "packages/engine.io": {
-      "version": "6.6.3",
+      "version": "6.6.4",
       "license": "MIT",
       "dependencies": {
         "@types/cors": "^2.8.12",
--- a/packages/socket.io/lib/index.ts
+++ b/packages/socket.io/lib/index.ts
@@ -821,7 +821,17 @@ export class Server<
     restoreAdapter();
 
     if (this.httpServer) {
-      this.httpServer.close(fn);
+      return new Promise<void>((resolve, reject) => {
+        this.httpServer.close((err) => {
+          if (err) {
+            fn && fn(err);
+            reject(err);
+          } else {
+            fn && fn();
+            resolve();
+          }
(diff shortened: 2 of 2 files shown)


package-lock.json
```diff
@@ -15507,7 +15507,7 @@
       }
     },
     "packages/engine.io": {
-      "version": "6.6.3",
+      "version": "6.6.4",
```
fantua, line 15510: related to https://github.com/socketio/socket.io/commit/72d61dab822a2f932b828bbfdc500a566c7cc87f

piotr-cz:
I think it's good moment to deprecate the error argument here:

https://github.com/socketio/socket.io/blob/91e1c8b3584054db6072046404a24e79a17c1367/packages/socket.io/lib/index.ts#L807

darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/bb0b480d2ab3108a8ae255b539015da451fdb249, and included in `socket.io@4.8.2`. Thanks!

@piotr-cz we will do this in the next major release

Related: https://github.com/socketio/socket.io/issues/4546

piotr-cz:
@darrachequesne 

How about deprecating the callback (docblock, docs) in minor/ patch version and removing in major version?

Issue #5353, opened 2025-06-02
Pooled memory management currently impossible with socket.io interface


bytenik:
**Is your feature request related to a problem? Please describe.**
Currently, my application maintains reasonably large Buffer objects in a pool, reusing them to avoid tons of GC contention. However, socket.io is a bit of a black hole for those buffers. Once we `.emit` one, we don't know when it has been actually sent to the client and "released" from socket.io. Thus we cannot reuse those buffers and they must be GC.

**Describe the solution you'd like**
I would like socket.io to expose a `released` event that is emitted every time socket.io has successfully sent an object to the client. It would contain as its body the object that was released. These could be buffers, classes,  POJO, etc., anything that lives on the heap and socket.io is now done with.

**Describe alternatives you've considered**
I don't see any alternatives at this time.

**Additional context**
I'd be very happy to implement this feature myself but before I spent time on it, I want to know that the solution I proposed is acceptable and the PR would likely to be merged.

bytenik:
@darrachequesne is this something you would be open to?

darrachequesne:
Hi!

Could the `drain` event of the low-level connection fit your use case?

```js
io.on("connection", (socket) => {
  socket.conn.on("drain", () => {
    // clean up
  });
});
```

Reference: https://socket.io/docs/v4/server-api/#socketconn

bytenik:
@darrachequesne No; that happens when the low-level buffer that's internalized to the socket is drained. It doesn't happen when _socket.io_ is done with the buffer because it has been placed into that TCP buffer.

bytenik:
@darrachequesne sorry to ping again and poke, but this is a huge issue at my company and I'm mostly just looking for your acknowledgement that you find my approach acceptable. At that point I'll start working on it. I would rather not develop something that then turns into a patch I have to apply to socket.io forever on my server.

darrachequesne:
Yes, that sounds like a reasonable use case 👍 could you please open a PR?

bytenik:
Hi @darrachequesne, the PR has been opened now with no engagement for several months. Can we do something to resolve this and get it merged?

darrachequesne:
Hi!

Sorry for the delay. To be honest, I think the suggested change needlessly complicates the source code, for a really specific use case, which is why I'm not sure about merging it.

PR #5362, opened 2025-06-17
Added argument to drain that provides the list of packets that are drained

bytenik:

* [ ] a bug fix
* [x] a new feature
* [ ] other

### Current behavior
`'drain'` event occurs with no arguments

### New behavior
`'drain'` event occurs with an argument containing the list of packets that were sent by the transport


darrachequesne:
Hmm, shouldn't you go a bit deeper and catch the `drain` event emitted by the HTTP response? In order to be sure that the packet has been written to the kernel buffer and can safely be released.

References:

- https://nodejs.org/api/http.html#responsewritechunk-encoding-callback
- https://nodejs.org/api/http.html#event-drain

Related: https://github.com/socketio/socket.io/issues/5353

bytenik:
> Hmm, shouldn't you go a bit deeper and catch the `drain` event emitted by the HTTP response? In order to be sure that the packet has been written to the kernel buffer and can safely be released.
> 
> References:
> 
> * https://nodejs.org/api/http.html#responsewritechunk-encoding-callback
> * https://nodejs.org/api/http.html#event-drain
> 
> Related: #5353

For Websockets, the ws library is handling that I think and it won't callback until its finished writing out. Do you see something different?

You are possibly right about long-polling though; I can defer our drain until the HTTP drain. I just pushed out that change. Though does long-polling even support buffers? I was under the impression it was base64-ing them.

I am not seeing a mechanism for uWebSockets to tell us when they have actually written out the buffer.

bytenik:
Hi, just wanted to check in on this PR @darrachequesne; I can update it to eliminate the conflicts but I don't want to keep doing it if it won't merge soon.

bytenik:
@darrachequesne Can I do anything to help get more engagement on this PR? Thanks!

codeCraft-Ritik (commented):
Nice enhancement! Providing packet details with the drain event improves observability and debugging.
This adds useful context without breaking existing behavior 👍

PR #5379, opened 2025-08-11
fix(socket.io): update socket.io-adapter@2.5.5

quinn-gaoo:

* [] a bug fix
* [ ] a new feature
* [x] other

### Current behavior
The dependencies in the package.json file of socket.io are outdated.

### New behavior
This PR updates the versions of the dependencies in the package.json file to the latest stable releases.

No related issues.
This PR does not introduce any breaking changes or modify the application logic; it only updates dependency versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15507,7 +15507,7 @@
       }
     },
     "packages/engine.io": {
-      "version": "6.6.3",
+      "version": "6.6.4",
       "license": "MIT",
       "dependencies": {
         "@types/cors": "^2.8.12",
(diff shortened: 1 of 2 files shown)


darrachequesne:

However, I don't think this is needed, as we are using tilde version range (`"engine.io": "~6.6.0"`), which automatically installs the latest patch release (i.e. `engine.io@6.6.4` with `~6.6.0`).

Reference: https://docs.npmjs.com/about-semantic-versioning

quinn-gaoo:
@darrachequesne Thanks for the explanation! I wasn't aware that ~6.6.0 already covers the latest patch updates. I'll close this PR since it's not needed.

darrachequesne:
No problem, have a great day!

PR #5387, opened 2025-09-02
drop CJS, drop IE, rebaseline platforms, enforce strict TS

sakgoyal:

* [x] a bug fix
* [ ] a new feature
* [x] other

### New behavior
* Drops support for CJS formats (ESM only. comon guys. ESM is everywhere already)
* still requiring IE11 support?
* all platforms have a new baseline set of features available everywhere (like Buffer/Blob)
* correctly set up typescript for modern platforms
	* make TS checking more strict to find more bugs

(WIP)


darrachequesne:

However, I don't think this is a sensible change, given the current state of the Node.js ecosystem. In a few years maybe?

sakgoyal:
what do you mean by the current state? 

node as far back as 16 has ESM support last I checked. 16 is very old now and has been long deprecated.

PR #5442, opened 2026-01-10
fix: cleanup pending acks on timeout to prevent memory leak

### LINKED ISSUE #4984: "emitWithAck" function leaks memory when acknowledgement timeout
**Describe the bug**

When using "emitWithAck" on the server side, the memory is not freed when the acknowledgement timeout is reached.

I tested on 500 client connections, and the memory usage is 1557MB after 3 hours.

Steps:

1. Connect 500 clients, 10 clients per room, 50 rooms
2. The first client in each room sends a message to the room every second(use emitWithAck)
3. Each client does not call callback after receiving a message(simulation timeout)
4. Memory usage is not freed

**To Reproduce**

*Server*

```js
try{
    const responses = await socket.timeout(2000).to("room1").emitWithAck("event", "message");
} catch(err){

}
```

Socket.IO client version: `4.7.2`

*Client*

```js
 socket.on("event", (data, callback) => {
      // Do not call callback, simulation timeout.
      // callback(1);
    });
```

**Expected behavior**

Should not leak memory

**Platform:**
 - Server Node.js version: `18.18.0`
 - Client Node.js version: `18.18.0`
 - OS: `win 10`

**Additional context**

See  [broadcast-operator.ts timer](https://github.com/socketio/socket.io/blob/b0568b20c13017d09c6565905a5c4cfa8d19ee2b/lib/broadcast-operator.ts#L237), it seems that when timeout, ack function is not deleted from [socket.acks](https://github.com/socketio/socket.io-adapter/blob/5eae5a0b541b2d671b561e88029f8aa888cb33f9/lib/in-memory-adapter.ts#L221).

seungeonchoi:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

When using `emitWithAck` with a timeout, if clients don't respond and the timeout triggers, the ack callbacks remain in `socket.acks` Map indefinitely, causing a memory leak.

### New behavior

The ack callbacks are cleaned up from `socket.acks` Map when the timeout occurs, preventing memory leak.

Fixes #4984

### DIFF
--- a/packages/socket.io-adapter/lib/cluster-adapter.ts
+++ b/packages/socket.io-adapter/lib/cluster-adapter.ts
@@ -482,7 +482,7 @@ export abstract class ClusterAdapter extends Adapter {
     opts: BroadcastOptions,
     clientCountCallback: (clientCount: number) => void,
     ack: (...args: any[]) => void,
-  ) {
+  ): { cleanup: () => void } {
     const onlyLocal = opts?.flags?.local;
     if (!onlyLocal) {
       const requestId = randomId();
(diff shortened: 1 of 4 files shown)


darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/da04267ffc7b0903ca91f2fccb80e56246d13328. Thanks a lot :+1:

PR #5452, opened 2026-02-14
fix: clean up socket.acks on broadcastWithAck timeout

### LINKED ISSUE #4984: "emitWithAck" function leaks memory when acknowledgement timeout

veeceey:
Fixes #4984

When using `emitWithAck` with a timeout on broadcast operations (e.g. `io.timeout(2000).to("room").emitWithAck("event", data)`), the ack callbacks stored in each socket's `acks` map were never removed if the client didn't respond before the timeout.

This causes a memory leak because `socket.acks.set(packet.id, ack)` is called in `broadcastWithAck` for every target socket, but when the timeout fires in `broadcast-operator.ts`, only the caller-side ack gets invoked - the per-socket entries in `socket.acks` are left behind indefinitely.

The fix tracks which sockets received the broadcast ack and schedules cleanup of their `acks` entries after the timeout period. This matches the existing pattern in `registerAckCallback` on individual sockets, which already cleans up on timeout (line 324 of socket.ts).

The issue is straightforward to reproduce: broadcast `emitWithAck` to a room where clients never call the callback, and observe that `socket.acks` grows unboundedly. The reporter measured 1.5 GB after 3 hours with 500 clients.

Includes unit tests verifying:
- ack entries are cleaned up after timeout expires
- no cleanup timer is set when timeout is not specified

### DIFF
--- a/packages/socket.io-adapter/lib/in-memory-adapter.ts
+++ b/packages/socket.io-adapter/lib/in-memory-adapter.ts
@@ -214,12 +214,14 @@ export class Adapter extends EventEmitter {
     const encodedPackets = this._encode(packet, packetOpts);
 
     let clientCount = 0;
+    const sockets: any[] = [];
 
     this.apply(opts, (socket) => {
       // track the total number of acknowledgements that are expected
       clientCount++;
       // call the ack callback for each client response
       socket.acks.set(packet.id, ack);
+      sockets.push(socket);
 
       if (typeof socket.notifyOutgoingListeners === "function") {
         socket.notifyOutgoingListeners(packet);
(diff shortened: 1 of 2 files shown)


veeceey:
quick bump on this — lmk if there's anything I can improve

PR #5453, opened 2026-02-15
fix(types): allow emitWithAck for events with void callbacks

### LINKED ISSUE #5257: Server emitWithAck on typed Server has "never" typing
**Describe the bug**

`emitWithAck` doesn't seem to be picking up types correctly, so the type variable `Ev` is assigned `never`.

**To Reproduce**

Socket.IO server version: `4.8.1`

*Server*

```typescript

interface ServerToClientEvents {
  someEvent: (callback: (err: Error) => void) => void;
}
interface ClientToServerEvents {}

const io = new Server<ClientToServerEvents, ServerToClientEvents>();
io.on("connection", async (socket) => {
  // Typescript gives error "Argument of type 'string' is not assignable to parameter of type 'never'.ts(2345)"
  await socket.emitWithAck("someEvent");

  // No error
  socket.emit("someEvent", (err) => {});
});
```

*Client*

No client code necessary.

**Expected behavior**
No Typescript error (similar to `emit` behavior)

**Platform:**
 - Device: Macbook Air M2
 - OS: MacOS Sonoma 14.6.1

**Additional context**
Seems to be the same bug as mentioned in the comment https://github.com/socketio/socket.io/issues/4925#issuecomment-2198057978, but different from that issue.

veeceey:
Fixes #5257

`EventNamesWithAck` was filtering out events whose acknowledgement callback had no non-error arguments (like `(cb: () => void) => void` or `(cb: (err: Error) => void) => void`). This meant you couldn't use `emitWithAck` as a simple ack mechanism without passing data back — even though `emit` worked fine with those same event signatures.

The root cause was the `FirstNonErrorArg<...> extends void ? never : K` check. For callbacks like `() => void`, `FirstNonErrorArg` resolves to `undefined` (since there are no params), and `undefined extends void` is true in TypeScript, so the event name got mapped to `never`.

The fix removes that void exclusion while keeping the guard for events with no parameters at all (`Parameters<Map[K]> extends never[]`), so events like `noArgs: () => void` are still correctly excluded from `emitWithAck`.

**Before:**
```ts
interface ServerToClientEvents {
  someEvent: (callback: (err: Error) => void) => void;
  ackOnly: (callback: () => void) => void;
}

// Both give: Argument of type 'string' is not assignable to parameter of type 'never'
await socket.emitWithAck("someEvent");
await socket.emitWithAck("ackOnly");
```

**After:**
```ts
// Both work correctly, returning Promise<undefined>
await socket.emitWithAck("someEvent");
await socket.emitWithAck("ackOnly");
```

All existing type tests and unit tests pass.

### DIFF
--- a/packages/socket.io/lib/typed-events.ts
+++ b/packages/socket.io/lib/typed-events.ts
@@ -22,8 +22,6 @@ export type EventNames<Map extends EventsMap> = keyof Map & (string | symbol);
 
 /**
  * Returns a union type containing all the keys of an event map that have an acknowledgement callback.
- *
- * That also have *some* data coming in.
  */
 export type EventNamesWithAck<
   Map extends EventsMap,
@@ -32,11 +30,11 @@ export type EventNamesWithAck<
   Last<Parameters<Map[K]>> | Map[K],
   K,
   K extends (
-    Last<Parameters<Map[K]>> extends (...args: any[]) => any
-      ? FirstNonErrorArg<Last<Parameters<Map[K]>>> extends void
-        ? never
-        : K
-      : never
+    Parameters<Map[K]> extends never[]
+      ? never
+      : Last<Parameters<Map[K]>> extends (...args: any[]) => any
+        ? K
+        : never
   )
     ? K
     : never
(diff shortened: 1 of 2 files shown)


veeceey:
hey, would appreciate a review when someone has a moment — happy to address any feedback

darrachequesne:
Awesome, thanks :+1:

Issue #5456, opened 2026-02-20
ClusterAdapter.broadcast() silently drops events to local sockets when publishAndReturnOffset throws error


vishnukumarya-kore:
### Description

  In `ClusterAdapter.broadcast()` (`cluster-adapter.js`, line 242–261), when
  `publishAndReturnOffset()` rejects (e.g. due to a serialization error in the
  adapter layer), the `catch` block does a **`return debug(...)`** which
  prevents `super.broadcast(packet, opts)` from ever being called.

  This means sockets connected to the **same server instance** that emitted the
  event also never receive it — a completely silent failure.

  ### Affected code

  ```js
  // cluster-adapter.js lines 242–261
  async broadcast(packet, opts) {
      const onlyLocal = opts?.flags?.local;
      if (!onlyLocal) {
          try {
              const offset = await this.publishAndReturnOffset({
                  type: MessageType.BROADCAST,
                  data: { packet, opts: encodeOptions(opts) },
              });
              this.addOffsetIfNecessary(packet, opts, offset);
          } catch (e) {
              return debug("[%s] error while broadcasting message: %s", this.uid, e.message);
              // ^^^ `return` here means super.broadcast() below is NEVER reached
          }
      }
      super.broadcast(packet, opts);  // ← never called when catch fires
  }
  ```

  Steps to reproduce

  1. Use createShardedAdapter (Redis 7+ sharded pub/sub) with an ioredis Cluster.
  2. Emit an event where the payload causes encode() inside doPublish() to throw
  (e.g. a Mongoose document — see related issue in [@socket.io/redis-adapter](https://github.com/socketio/socket.io-redis-adapter/issues/572)).
  3. Observe that no connected socket receives the event, including sockets on
  the same Node.js process.

  Expected behaviour

  When remote publishing fails, the error should be logged but super.broadcast() should still be called so that local sockets on the same server instance receive the event.

  Environment

  - socket.io-adapter: 2.x (ClusterAdapter / ClusterAdapterWithHeartbeat)
  - @socket.io/redis-adapter: 8.3.x (ShardedRedisAdapter)
  - Redis: 7.x cluster with sharded pub/sub (SPUBLISH/SSUBSCRIBE)
  - Node.js: 18.x

idirdev:
[...]

PR #5457, opened 2026-02-21
fix(adapter): do not skip local broadcast when publishAndReturnOffset throws

Not-Sarthak:
## Summary

Fixes https://github.com/socketio/socket.io/issues/5456

In `ClusterAdapter.broadcast()`, the `catch` block used `return debug(...)` which prevented `super.broadcast(packet, opts)` from ever being called when `publishAndReturnOffset()` rejected. This meant sockets connected to the **same server instance** that emitted the event would silently never receive it.

The fix removes the `return` keyword so that `super.broadcast()` is still called after logging the error — matching the pattern already used in `addSockets`, `delSockets`, and `disconnectSockets`.

- **Before:** remote publish error → `return debug(...)` → `super.broadcast()` skipped → local sockets get nothing
- **After:** remote publish error → `debug(...)` → `super.broadcast()` still runs → local sockets receive the event

## Test plan

- [x] Added test case: `"broadcasts to local clients even when publishAndReturnOffset throws"`
- [x] All 48 existing adapter tests still pass
- [x] `engine.io` tests pass (189/189)
- [x] `engine.io-client` tests pass (93/93)

### DIFF
--- a/packages/socket.io-adapter/lib/cluster-adapter.ts
+++ b/packages/socket.io-adapter/lib/cluster-adapter.ts
@@ -438,11 +438,7 @@ export abstract class ClusterAdapter extends Adapter {
         });
         this.addOffsetIfNecessary(packet, opts, offset);
       } catch (e) {
-        return debug(
-          "[%s] error while broadcasting message: %s",
-          this.uid,
-          e.message,
-        );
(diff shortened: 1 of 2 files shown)


darrachequesne:
Thanks :+1:

Issue #5250, opened 2024-12-04
Event triggered on Connection State Recovery Expiration


bryghtlabs-richard:
**Is your feature request related to a problem? Please describe.**

Per-connection, we're keeping a large object in memory, which we would prefer to use across connections that are recovered through connection-state-recovery. Connection state recovery solves one problem for us, in that we can reconnect a client to its object, but this makes it harder to know when to delete the object.

**Describe the solution you'd like**

When a connection exceeds the connection state recovery maxDisconnectionDuration, I'd like an event trigger with the SocketID as an argument.

**Describe alternatives you've considered**

We've considered setting an additional timeout, but this won't ever be as accurate - we'd have to make this timeout longer than maxDisconnectionDuration, and deal with our large staying around longer than needed.

darrachequesne:
Hi! That's an interesting idea :+1: 

I think implementing this for the in-memory adapter would be quite straightforward.

For the adapter based on Redis streams, we would need to listen to the "expired" events ([ref](https://redis.io/docs/latest/develop/use/keyspace-notifications/)), though it would not be reliable if the Redis client disconnects.

Not sure how we could implement it for the adapter based on MongoDB though...

Regarding the API, would something like this suit your needs?

```js
io.of("/").adapter.on("sessionExpiry", (sid) => {
  // ...
});
```

PR #5478, opened 2026-03-19
feat: emit a "session-expiry" event when a session has expired

zaewc:

* [ ] a bug fix
* [x] a new feature
* [ ] other

### Current behavior

When a session expires and is removed in SessionAwareAdapter, no event is emitted, making it impossible for external consumers to detect when a session has expired.  

### New behavior

SessionAwareAdapter now emits a "session-expiry" event whenever an expired session is deleted. The event handler receives the sid of the expired session as an argument.  

https://github.com/socketio/socket.io/issues/5250

### DIFF
--- a/packages/socket.io-adapter/lib/in-memory-adapter.ts
+++ b/packages/socket.io-adapter/lib/in-memory-adapter.ts
@@ -423,6 +423,7 @@ export class SessionAwareAdapter extends Adapter {
         const hasExpired = session.disconnectedAt < threshold;
         if (hasExpired) {
           this.sessions.delete(sessionId);
+          this.emit("session-expiry", session.sid);
         }
       });
       for (let i = this.packets.length - 1; i >= 0; i--) {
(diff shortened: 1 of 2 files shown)


darrachequesne:

Could you please explain your use case for this feature?

zaewc:
@darrachequesne 

  Sure!! When using connection state recovery, there's currently no way to know when a persisted session has expired. This makes it difficult to perform cleanup logic tied to session lifetime                                                                  
                                                                                                                                  
  e.g.                                                                                                                    
  - Updating a user's online/presence status when their session can no longer be recovered                                      
  - Cleaning up server side resources like cached user data that were kept alive in anticipation of reconnection               
  - Logging or analytics around session recovery failures                                                          
                                                                                                                      
            
  Without this event, the only option is to poll or rely on external timeouts, which are harder to synchronize with the actual expiry logic inside SessionAwareAdapter.

codeCraft-Ritik (commented):
Great improvement! Emitting a session-expiry event makes session lifecycle changes observable and easier to handle.
Nice enhancement for extensibility and integration 👍

PR #5482, opened 2026-03-31
fix(engine.io): make _query optional in EngineRequest

Akash9541:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

The `_query` field in `EngineRequest` is typed as required (`Record<string, string>`), while in practice it is initialized later in the `prepare()` method. This creates a mismatch between the TypeScript definition and the runtime behavior, leading to type errors in certain environments (e.g. when following the Nuxt integration tutorial).

### New behavior

The `_query` field is now marked as optional in the `EngineRequest` type to accurately reflect its initialization lifecycle.

Additionally, a runtime check ensures that `_query` is defined before being passed to the Transport constructor, preserving type safety where `_query` is expected to be available.

Related to the discussion about `_query` initialization and type mismatch when using Engine.IO with certain frameworks (e.g. Nuxt).

### DIFF
--- a/packages/engine.io/lib/transport.ts
+++ b/packages/engine.io/lib/transport.ts
@@ -13,7 +13,7 @@ function noop() {}
 type ReadyState = "open" | "closing" | "closed";
 
 export type EngineRequest = IncomingMessage & {
-  _query: Record<string, string>;
+  _query?: Record<string, string>;
   res?: ServerResponse;
   cleanup?: Function;
   websocket?: WebSocket & {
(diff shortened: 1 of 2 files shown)


Akash9541:
@darrachequesne can you please check it?

darrachequesne:
Hi! Thank you for your pull request.

Upon review, I think that using `IncomingMessage` instead of `EngineRequest` in the exposed API (`handleRequest()`, `handleUpgrade()`) is cleaner in that case: https://github.com/socketio/socket.io/commit/b059af6b12e69729540e78751e9be080fd1e7fb5

For reference, the `EngineRequest` type was introduced in https://github.com/socketio/socket.io/commit/c310b7b6b66236f312f017583a67f81389fa5d72.

Thanks again!

PR #5495, opened 2026-05-01
fix(eio): make _query optional in EngineRequest

### LINKED ISSUE #5468: [engine.io] `_query` should be optional
Hi!

Following the Nuxt tutorial (https://socket.io/how-to/use-with-nuxt) and as a follow-up on this ( #5323 ) unanswered discussion, I think the following line/variable `_query` should be made optional:

https://github.com/socketio/socket.io/blob/07cbe1510ded7e5460cb82e026e2533e50e30eaf/packages/engine.io/lib/transport.ts#L16

Since `handleRequest()` in:

https://github.com/socketio/socket.io/blob/07cbe1510ded7e5460cb82e026e2533e50e30eaf/packages/engine.io/lib/server.ts#L758

calls `prepare()`, which sets the `_query` field even if it doesn't exist yet:

https://github.com/socketio/socket.io/blob/07cbe1510ded7e5460cb82e026e2533e50e30eaf/packages/engine.io/lib/server.ts#L734-L740

And my editor currently complains about the types mismatch after completing the above mentionned tutorial:

Image

Thanks in advance!

elementnl:
Fixes #5468

`EngineRequest` declared `_query` as required, but the property is set lazily by `prepare()` and doesn't exist on a raw `IncomingMessage`. This caused a TypeScript error when passing a plain `IncomingMessage` where an `EngineRequest` is expected; for example, in Nuxt integrations following the official guide.

Before: Any code that worked with `EngineRequest` directly would get a TypeScript mismatch error because `_query` was required but not present on the underlying `IncomingMessage`.

After: `_query` is typed as optional in `EngineRequest` and in the `Transport` constructor, which correctly reflects the lifecycle of the property: it may not be set at construction time, but is always guaranteed to be present by the time any transport logic runs. `transports/index.ts` is also updated for consistency.

No tests added - this is a TypeScript-only change with no runtime behavior difference.

### DIFF
--- a/packages/engine.io/lib/transport.ts
+++ b/packages/engine.io/lib/transport.ts
@@ -13,7 +13,7 @@ function noop() {}
 type ReadyState = "open" | "closing" | "closed";
 
 export type EngineRequest = IncomingMessage & {
-  _query: Record<string, string>;
+  _query?: Record<string, string>;
   res?: ServerResponse;
   cleanup?: Function;
   websocket?: WebSocket & {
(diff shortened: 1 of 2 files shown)


darrachequesne:
Thanks for your report :+1: 

I think this was (indirectly) fixed by https://github.com/socketio/socket.io/commit/b059af6b12e69729540e78751e9be080fd1e7fb5, included in `engine.io@6.6.7`. The public API now uses a plain `IncomingMessage` object.

So I think we can close this. Please reopen if needed.

PR #5499, opened 2026-05-17
fix(engine.io-client): include query params in WebTransport URI

### LINKED ISSUE #5496: WebTransport ignores query parameters in URL
**Describe the bug**
The final WebTransport URL doesn't include the query parameters from the original URL.

**To Reproduce**

engine.io-client 6.6.4

*Client*

```js
// Connects to: https://localhost:1234/engine.io/?x=y&EIO=4&transport=polling&t=zsyh7czl 
eio("https://localhost:1234/?x=y", {transports: ["polling"]});

// Connects to: wss://localhost:1234/engine.io/?x=y&EIO=4&transport=websocket
eio("https://localhost:1234/?x=y", {transports: ["websocket"]});

// Connects to: https://localhost:1234/engine.io/
eio("https://localhost:1234/?x=y", {transports: ["webtransport"]});
```

**Expected behavior**
The WebTransport transport should preserve the query parameters in the same way other transports do.

**Additional context**

WebTransport currently uses `createUri` to construct the URL:

https://github.com/socketio/socket.io/blob/439a8f669c6788b512dd76011c4476830d67515b/packages/engine.io-client/lib/transports/webtransport.ts#L31-L34

It doesn't include the original query parameters:

https://github.com/socketio/socket.io/blob/439a8f669c6788b512dd76011c4476830d67515b/packages/engine.io-client/lib/transport.ts#L176-L185

cyphercodes:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

The WebTransport client creates its session URL without passing the transport query object, so query parameters from the original Engine.IO URL are dropped.

### New behavior

The WebTransport client now passes the same query object used by the other transports when constructing its URL, preserving original query parameters along with Engine.IO transport metadata.

Fixes #5496.

Validation:

- `npm run format:check --workspace=engine.io-client`
- `npm run compile --workspace=engine.io-parser && npm run compile --workspace=engine.io-client`
- `npx mocha --bail --require test/support/hooks.js test/transport.js`
- `npm run test:node --workspace=engine.io-client -- --grep "autoRef option" --timeout 10000`
- `git diff --check`

Note: `npm test --workspace=engine.io-client` timed out in the existing `node.js > autoRef option > should stop once the timer is triggered` test with the default 2s timeout on Node v22 in this environment; the isolated autoRef test group passed with a higher timeout.

### DIFF
--- a/packages/engine.io-client/lib/transports/webtransport.ts
+++ b/packages/engine.io-client/lib/transports/webtransport.ts
@@ -29,7 +29,7 @@ export class WT extends Transport {
     try {
       // @ts-ignore
       this._transport = new WebTransport(
-        this.createUri("https"),
+        this.createUri("https", this.query),
         this.opts.transportOptions[this.name],
       );
     } catch (err) {
--- a/packages/engine.io-client/test/transport.js
+++ b/packages/engine.io-client/test/transport.js
@@ -198,6 +198,35 @@ describe("Transport", () => {
       expect(ws.uri()).to.be("wss://test/engine.io");
     });
 
+    it("should include query parameters in WebTransport uris", () => {
+      let actualUri;
+      const previousWebTransport = globalThis.WebTransport;
+      globalThis.WebTransport = function (uri) {
+        actualUri = uri;
+        this.closed = new Promise(() => {});
+        this.ready = new Promise(() => {});
+      };
+
+      try {
+        new eio.transports.webtransport({
+          path: "/engine.io",
+          hostname: "test",
+          secure: true,
+          query: { token: "abc", EIO: "4", transport: "webtransport" },
+          transportOptions: { webtransport: {} },
+        }).open();
+      } finally {
+        globalThis.WebTransport = previousWebTransport;
+      }
+
+      const url = new URL(actualUri);
+      expect(url.origin).to.be("https://test");
+      expect(url.pathname).to.be("/engine.io");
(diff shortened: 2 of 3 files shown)


darrachequesne:
Unfortunately, I think this would require to update the server implementation too, because the path does not match there (because the query params must be stripped): https://github.com/socketio/socket.io/blob/1e2ae34351908080d0c81ca397121027ed0c779c/packages/engine.io-client/test/webtransport.mjs#L34

We would have to add something like:

```js
h3Server.setRequestCallback(async ({ header }) => {
  const url = new URL(header[":path"], "http://localhost");
  if (url.pathname === "/engine.io/") {
    // ...
  }
  return { status: 404 };
});
```

cyphercodes:
Updated the WebTransport test harness to match the session path by pathname so query parameters in the client URI do not prevent the test HTTP/3 server from accepting `/engine.io/`.

Local verification:
- `npx prettier --check 'packages/engine.io-client/test/webtransport.mjs'` ✅
- `npm run test:node -w engine.io-client` ✅ (`94 passing`, `1 pending`)
- `git diff --check` ✅

darrachequesne:
Thanks for the update. However, I think this is a breaking change which would disrupt existing WebTransport connections, as the use of `setRequestCallback()`  would now be required.

Possibly related: https://github.com/socketio/socket.io/commit/d1f5aa93722a7f1ed729b96f771daf92a3dfdaf7

PR #5501, opened 2026-05-23
fix(engine.io-client): preserve caller transport options

### LINKED ISSUE #5462: socket.io-client - transports options array modified from [string] to [object] by io constructor
**Describe the bug**

```
const options = {transports: ["websocket","polling"]};
const s = io(<URL>, options);
```
options object is modified, and transports is set to  [ class WS, class XHR ]
Not really a big issue, but quite unexpected and an unusual behaviour. Especially when you need to serialize the options object after connection.

**Expected behavior**
options object should not be mutated.

puneetdixit200:
## Summary
- clone Socket constructor options before mapping transport names to constructors
- pass the cloned options through the options-only overload as well
- add a regression test for both URI and options-only construction

Fixes #5462

## Tests
- `npm run compile --workspace=engine.io-parser`
- `npm run compile --workspace=engine.io-client`
- `npm run format:check --workspace=engine.io-client`
- `npm run test:node --workspace=engine.io-client -- --grep "caller's transports option"`
- `npm run test:node --workspace=engine.io-client -- --grep "Socket"`
- `npm run test:node --workspace=engine.io-client`
- `git diff --check`

### DIFF
--- a/packages/engine.io-client/lib/socket.ts
+++ b/packages/engine.io-client/lib/socket.ts
@@ -1167,7 +1167,8 @@ export class Socket extends SocketWithUpgrade {
   constructor(uri?: string, opts?: SocketOptions);
   constructor(opts: SocketOptions);
   constructor(uri?: string | SocketOptions, opts: SocketOptions = {}) {
-    const o = typeof uri === "object" ? uri : opts;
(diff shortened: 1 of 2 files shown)


darrachequesne:
Great, thanks :+1:

PR #5519, opened 2026-07-02
fix(sio): reset regex lastIndex when matching dynamic namespaces

spokodev:
A dynamic namespace registered with a `RegExp` that carries the global (`g`) or
sticky (`y`) flag rejects connections intermittently.

`Server.of` and the parent-namespace matcher call `.test()` on the user supplied
regex as if it were a stateless predicate. Per the ECMAScript spec,
`RegExp.prototype.test()` advances `lastIndex` when the regex has the `g` or `y`
flag and resumes from it on the next call, so a shared regex object returns
alternating results across connections.

Concretely, `io.of(/^\/room-\d+$/g)` connects `/room-1`, then rejects `/room-2`
with `connect_error: "Invalid namespace"`, then connects `/room-3`, and so on:

    const re = /^\/room-\d+$/g
    re.test('/room-1') // true
    re.test('/room-2') // false (wrong)
    re.test('/room-3') // true

The pattern is correct; only the leftover `lastIndex` state causes the rejection.

The fix resets `lastIndex` to 0 immediately before each `.test()` at both call
sites. It is a no-op for regexes without the `g` or `y` flag, so existing
behavior is unchanged.

### DIFF
--- a/packages/socket.io/lib/index.ts
+++ b/packages/socket.io/lib/index.ts
@@ -788,10 +788,12 @@ export class Server<
       if (typeof name === "function") {
         this.parentNsps.set(name, parentNsp);
       } else {
-        this.parentNsps.set(
-          (nsp, conn, next) => next(null, (name as RegExp).test(nsp)),
-          parentNsp,
-        );
+        this.parentNsps.set((nsp, conn, next) => {
+          // Reset lastIndex so a regex created with the global or sticky flag
+          // matches consistently across connections.
+          (name as RegExp).lastIndex = 0;
+          next(null, (name as RegExp).test(nsp));
+        }, parentNsp);
         this.parentNamespacesFromRegExp.set(name, parentNsp);
       }
       if (fn) {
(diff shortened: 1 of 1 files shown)


darrachequesne:
@spokodev hi! Could you please explain your use case regarding global regular expressions?

spokodev:
Fair question. I am not deliberately setting a global flag on a namespace regex; the value here is defensive. `Server.of` accepts whatever RegExp the caller passes, and if that object carries `g` or `y` (a shared or reused regex, one produced by a helper or config, or a copy-pasted literal), `.test()` advances its `lastIndex` and the namespace then matches only every other connection. That is a silent, order-dependent failure that is hard to trace back to the flag. Resetting `lastIndex` before matching keeps the result independent of flags the caller did not intend to affect namespace routing. If you would rather guard against it another way (for example rejecting global/sticky regexes at registration), I can adjust.

darrachequesne:
Merged as https://github.com/socketio/socket.io/commit/59f4f2413b2cd1d6a1cf91d367bac71af51f4a24.

Thanks for the heads-up!

PR #5523, opened 2026-07-13
fix(eio): do not close the connection if other packets are received while waiting for a ping/pong

### LINKED ISSUE #5450: Heartbeat timeout is no longer reset on packet
**Describe the bug**
We have a high volume of data incoming from our sockets and they get disconnected at PingInterval+Timeout s with a ping timeout because the server has not received a pong, **however** it has received other messages from these sockets.

Image

Note how some connections drop after ~40seconds and reconnect.

There was a fix for this in https://github.com/socketio/engine.io/commit/be7b4e7478132a9409603327b27d1aa1970dd1d9 which got reverted in https://github.com/socketio/socket.io/commit/5359bae683e2a25742bd4989d0355a8fc10d294e

While I understand the "wastefulness" of creating timers on every packet, I still think the point of the socket being very much alive if there is data coming from it is valid (even when the pong packages is coming through timely).

Is there a chance to revert the revert?

**To Reproduce**

Socket.IO server version: `4.x`

Send a lot of packages from the client, watch them disconnect after ~40 seconds even when data is coming from them.

**Expected behavior**
Connections should not be considered stale if there are still packets coming.

GiHoon1123:
## Summary

Fixes #5450.

The ping timeout timer only ever considered the one specific packet it was waiting for (the pong on the server side, the next ping on the client side), so a connection could be closed as unresponsive even while it was actively exchanging other packets - e.g. a high-throughput client whose pong happens to be delayed behind a backlog of its own outgoing messages, as described in the issue.

## Approach

Both `engine.io` (server) and `engine.io-client` now track the timestamp of the last packet received, of any type. When the ping timeout timer fires, instead of closing unconditionally, it checks whether a packet has arrived recently enough that the connection is clearly still alive, and if so, extends the wait (rescheduled for the remaining time) instead of giving up.

This needed to be applied on **both** sides to actually help: the server's own re-ping schedule is only resumed once it receives a pong, so if only the server were patched, an indefinitely-delayed pong would eventually cause the *client's own*, independent ping-timeout watchdog to give up instead (it also only resets on receiving a fresh ping) - just moving which side reports the disconnect, not fixing it. With both sides checking "has anything arrived lately" before giving up, a connection with ongoing traffic in either direction survives a delayed/lost specific ping or pong, as long as evidence of liveness keeps arriving.

### Why not just refresh the timer on every packet (the previous attempt)

A previous attempt at exactly this ([engine.io@be7b4e7](https://github.com/socketio/engine.io/commit/be7b4e7478132a9409603327b27d1aa1970dd1d9)) refreshed the ping timeout timer directly on every incoming packet, and was reverted ([socket.io@5359bae](https://github.com/socketio/socket.io/commit/5359bae683e2a25742bd4989d0355a8fc10d294e)) for two reasons, quoted from the revert commit message:

- "a new timer is allocated every time a packet is received, which is wasteful"
- "the next heartbeat is not actually delayed, since it's the timeout timer which gets reset, and not the interval timer... delaying the next heartbeat would be a breaking change"

This PR avoids both: tracking a plain timestamp on every packet is a cheap property write, not a timer allocation, on the common (nothing overdue) path. And the check only ever runs when a ping-timeout timer that's *already scheduled* fires - it never touches the interval timer that governs when the next ping is sent, so there's no change to the ping schedule itself, and no breaking change.

## Test plan

- Two new tests in `packages/engine.io/test/server.js`, one exercising the server-side check (with the client's own independent watchdog neutralized, to isolate the server's decision) and one exercising the client-side check (with the server's ping schedule stalled, to isolate the client's decision), both passing.
- Verified each new test actually catches a regression: with the corresponding fix reverted, the same test fails with a premature close at the expected time (confirmed via manual revert + re-run, not just inspection).
- Full existing suite passes, both protocol v4 (`npm run test:default`, 195 passing) and legacy protocol v3 (`npm run test:compat-v3`, 153 passing), no regressions.
- `engine.io-client`'s own test suite (`npm test`) passes, 94 passing.
- `npm run compile` and `npm run format:check` clean on both packages.

## Checklist

- [x] add one or more test cases, in order to avoid any regression in the future
- [x] make sure existing tests still pass

### DIFF
--- a/packages/engine.io-client/lib/socket.ts
+++ b/packages/engine.io-client/lib/socket.ts
@@ -348,6 +348,13 @@ export class SocketWithoutUpgrade extends Emitter<
    * callback is not fired on time. This can happen for example when a laptop is suspended or when a phone is locked.
    */
   private _pingTimeoutTime = Infinity;
+  /**
+   * Timestamp of the last packet received from the server, of any type. Used
+   * by the ping timeout check as a cheap liveness signal, so a server that is
+   * clearly still sending data isn't considered dead just because its next
+   * ping happens to be delayed (e.g. behind a backlog on its own end).
+   */
+  private _lastPacketTime: number = Date.now();
   private clearTimeoutFn: typeof clearTimeout;
   private readonly _beforeunloadEventListener: () => void;
   private readonly _offlineEventListener: () => void;
@@ -603,6 +610,7 @@ export class SocketWithoutUpgrade extends Emitter<
 
       // Socket is live - any packet counts
       this.emitReserved("heartbeat");
+      this._lastPacketTime = Date.now();
 
       switch (packet.type) {
         case "open":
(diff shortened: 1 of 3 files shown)


GiHoon1123:
Closing for now - want to take another look before this is ready. Apologies for the noise.

PR #5524, opened 2026-07-13
fix(eio,eio-client): do not close the connection while packets are still being received

### LINKED ISSUE #5450: Heartbeat timeout is no longer reset on packet

GiHoon1123:
## Summary

Fixes #5450.

The connection could be closed on a ping timeout even while other packets kept arriving, because only the pong (server side) or a fresh ping (client side) reset the timeout - a high-throughput connection could be dropped despite clear evidence it was still alive.

A previous fix for this ([be7b4e7](https://github.com/socketio/engine.io/commit/be7b4e7478132a9409603327b27d1aa1970dd1d9)) was reverted ([5359bae](https://github.com/socketio/socket.io/commit/5359bae683e2a25742bd4989d0355a8fc10d294e)) for two reasons: it allocated a new timer on every incoming packet, and it delayed the next scheduled ping as a side effect (since only the timeout timer was touched, not the interval timer).

This takes a different approach:
- A plain timestamp of the last received packet is tracked (updated on every packet, no timer touched on this path).
- Only when the pre-existing ping-timeout timer actually fires does a check run: if a packet arrived recently enough, the timer is extended by the remaining time instead of closing; otherwise the connection is closed as before.
- The ping/pong schedule itself (`pingIntervalTimer` on the server, the client's own ping-driven timer) is never touched, so there's no change to when pings are sent.

Both `engine.io` and `engine.io-client` needed the change. Delaying the server's wait for a pong also delays the server's next ping; without a matching client-side check, the client's own independent ping-timeout watchdog would still close the connection on its own once that delay exceeded `pingTimeout`. This was confirmed by testing the server-only fix in isolation - the client still disconnected on its own.

`_hasPingExpired()`, the synchronous check the Socket.IO client layer uses to detect a throttled/delayed timeout timer, is also updated to apply the same recent-packet check before deciding the connection is stale, since it's a separate path that can bypass the timer-based check entirely.

## Test plan

- [x] New test: server does not close the connection when the client's pongs are delayed but other packets keep arriving (client's own watchdog neutralized to isolate server-side behavior)
- [x] New test: client does not close the connection when the server's pings are delayed but other packets keep arriving (server's own ping suppressed to isolate client-side behavior)
- [x] New test: `_hasPingExpired()` returns `false` when a packet was recently received, even if the throttled-timer deadline has passed
- [x] Existing `engine.io` suite passes for both protocol v4 and legacy v3 (`EIO_CLIENT=3`)
- [x] Existing `engine.io-client` suite passes
- [x] Verified regression coverage by reverting each fix individually (`git stash` on just that file) and confirming the corresponding new test fails without it, then restoring and reconfirming it passes

### DIFF
--- a/packages/engine.io-client/lib/socket.ts
+++ b/packages/engine.io-client/lib/socket.ts
@@ -348,6 +348,13 @@ export class SocketWithoutUpgrade extends Emitter<
    * callback is not fired on time. This can happen for example when a laptop is suspended or when a phone is locked.
    */
   private _pingTimeoutTime = Infinity;
+  /**
+   * Timestamp of the last packet received from the server, of any type. Used
+   * by the ping timeout check as a cheap liveness signal, so a server that is
+   * clearly still sending data isn't considered dead just because its next
+   * ping happens to be delayed (e.g. behind a backlog on its own end).
+   */
+  private _lastPacketTime: number = Date.now();
   private clearTimeoutFn: typeof clearTimeout;
   private readonly _beforeunloadEventListener: () => void;
   private readonly _offlineEventListener: () => void;
@@ -603,6 +610,7 @@ export class SocketWithoutUpgrade extends Emitter<
 
       // Socket is live - any packet counts
       this.emitReserved("heartbeat");
+      this._lastPacketTime = Date.now();
 
       switch (packet.type) {
         case "open":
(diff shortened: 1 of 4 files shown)


GiHoon1123:
Just checking in on this one. No rush at all, and I'd be happy to make any updates if there's anything else needed from my side.

darrachequesne:
Superseded by https://github.com/socketio/socket.io/commit/da008a514ded71e3058a964c831c69abe417b2f5.

Thanks a lot for your work on this :+1:

PR #5525, opened 2026-07-13
fix(eio,eio-client): handle close packets consistently across transports

### LINKED ISSUE #5286: Inconsistent behavior of close packet processing in websocket and polling transports
**Describe the bug**
In the engine.io WebSocket transport, when a close packet is received, it is sent to the top in onPacket and the connection continues functioning as if nothing happened. In contrast, the polling transport closes the client and sends the close event to the top in onClose.

**To Reproduce**

Engine.IO server and client version: 72d61da Send close packet from server to client.

**Expected behavior**
All transports should handle data processing consistently.

GiHoon1123:
## Summary

Fixes #5286.

The polling transport already closes the connection when it receives a "close" packet, but the WebSocket transport silently ignored it and kept the connection running - on both the server and the client. This is inconsistent with the Engine.IO protocol, where `close` is a formal packet type (independent of the underlying WebSocket connection's own close frame), so any conformant implementation may send it over any transport.

## Approach

Added the same check to the shared base `Transport#onData()` in both `engine.io` and `engine.io-client`, rather than duplicating polling's own override:
- On the server, this also covers the uWebSockets.js-based WebSocket transport for free, since it relies on the same base method and doesn't override `onData()`.
- On the client, the close details now match the polling transport's existing wording (`"transport closed by the server"`), for consistency between transports.

Both fixes call the transport's real close procedure (`close()` / `doClose()` + `onClose()`), not just the internal state transition - an earlier version of this fix called `onClose()` directly (mirroring polling's own shortcut), but that leaves the underlying WebSocket connection open indefinitely, since polling's shortcut is safe only because each poll is a short-lived HTTP request that Node's HTTP server cleans up on its own, which doesn't apply to a persistent WebSocket connection.

## Out of scope

While investigating, two related gaps were found but intentionally left out of this PR to keep it focused on the reported issue:
- The `WebTransport` transport has the same underlying gap (it decodes and dispatches packets directly, bypassing `onData()` entirely), but it isn't mentioned in the issue and is disabled by default. Happy to follow up separately if maintainers want it addressed.
- This PR only reproduces the specific direction described in the issue (server sending a close packet, and the reverse). Nothing further intentionally deferred beyond `WebTransport` above.

## Test plan

- [x] New test: server closes the transport upon receiving a close packet from a WebSocket client
- [x] New test: client closes the transport upon receiving a close packet from the server (matches the issue's reproduction), including asserting the close details match polling's wording
- [x] Reverted each fix individually (`git stash` on just that file) and confirmed the corresponding new test fails without it, then restored and reconfirmed it passes
- [x] Checked directly (not just via the `readyState` field) that the underlying raw WebSocket connection is actually closed on both the server and client side, not just marked closed at the Engine.IO layer
- [x] Existing `engine.io` suite passes for both protocol v4 and legacy v3 (`EIO_CLIENT=3`) - the client-side test is skipped under legacy v3 mode, since that mode exercises the separately-published `engine.io-client-v3` package rather than the `engine.io-client` package touched by this PR
- [x] Existing `engine.io-client` suite passes

### DIFF
--- a/packages/engine.io-client/lib/transport.ts
+++ b/packages/engine.io-client/lib/transport.ts
@@ -139,6 +139,14 @@ export abstract class Transport extends Emitter<
    */
   protected onData(data: RawData) {
     const packet = decodePacket(data, this.socket.binaryType);
+    if ("close" === packet.type) {
+      debug("got close packet");
+      if (this.readyState === "opening" || this.readyState === "open") {
+        this.doClose();
+        this.onClose({ description: "transport closed by the server" });
+      }
+      return;
+    }
     this.onPacket(packet);
   }
 
--- a/packages/engine.io/lib/transport.ts
+++ b/packages/engine.io/lib/transport.ts
@@ -162,7 +162,13 @@ export abstract class Transport extends EventEmitter {
    * @protected
    */
   protected onData(data: RawData) {
-    this.onPacket(this.parser.decodePacket(data));
+    const packet = this.parser.decodePacket(data);
+    if ("close" === packet.type) {
+      debug("got close packet");
+      this.close();
+      return;
+    }
+    this.onPacket(packet);
   }
 
   /**
--- a/packages/engine.io/test/server.js
+++ b/packages/engine.io/test/server.js
@@ -1604,6 +1604,56 @@ describe("server", () => {
       });
     });
 
+    it("should close transport upon receiving a close packet (ws)", (done) => {
+      const opts = { allowUpgrades: false, transports: ["websocket"] };
+      const engine = listen(opts, (port) => {
+        engine.on("connection", (conn) => {
+          conn.transport.on("close", done);
+        });
+        const socket = new ClientSocket(`ws://localhost:${port}`, {
+          transports: ["websocket"],
+        });
+        socket.on("open", () => {
+          // bypass the client library entirely and send a raw close packet
+          // ("1", per the Engine.IO protocol), to make sure the server
+          // reacts the same way regardless of which client implementation
+          // is on the other end
+          socket.transport.ws.send("1");
+        });
+      });
+    });
+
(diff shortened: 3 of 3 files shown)


GiHoon1123:
Just checking in on this one. No rush at all, and I'd be happy to make any updates if there's anything else needed from my side.

packages/engine.io-client/lib/transport.ts
```diff
@@ -139,6 +139,14 @@ export abstract class Transport extends Emitter<
    */
   protected onData(data: RawData) {
     const packet = decodePacket(data, this.socket.binaryType);
+    if ("close" === packet.type) {
+      debug("got close packet");
+      if (this.readyState === "opening" || this.readyState === "open") {
+        this.doClose();
```
darrachequesne: I think ` onClose()` will be called twice here with WebSocket:

- ` this.onClose({ description: "transport closed by the server" });`
- and then in the "close" handler here: https://github.com/socketio/socket.io/blob/3df3fed075fd2312e84481786235aa14d08b8fe7/packages/engine.io-client/lib/transports/websocket.ts#L83-L86

Could you please check?

darrachequesne (changes requested):
Thanks for the pull request. Please find my review notes below.

GiHoon1123:
Thanks for catching this. I checked the WebSocket path and confirmed that the close packet called onClose() once, then ws.close() triggered the WebSocket onclose handler and called it again.

BaseWS.doClose() now clears that handler before closing the socket. The regression test counts Transport.onClose() calls and expects one; it failed with two calls before the change and passes now. The close reason and details checks remain in place.

The close-packet tests and format checks pass. The only local failure in the full Engine.IO run is the existing WebTransport test because the native webtransport.node module is not available in this environment.

PR #5528, opened 2026-07-16
fix(engine): support Node 10 query parsing

### LINKED ISSUE #5527: Using the function Object.fromEntries in a wrong node version
**Describe the bug**
I'm trying to use the lib with node 10.16.0 and inside the lib engine.io is using the function "Object.fromEntries" but it doesn't work in this version of node. This lib should work using this node because its package.json says it should work in node >=10.2.0.

tianrking:
## Summary

Fixes #5527.

Engine.IO advertises support for Node.js 10.2+, but both the standard HTTP server and the uWebSockets server parsed query strings with `Object.fromEntries()`, which was only added in Node.js 12. A polling handshake therefore threw before validation on supported Node 10 deployments.

Replace those calls with a shared `URLSearchParams` parser that uses `Object.defineProperty`. It preserves the previous last-value-wins behavior for duplicate keys and keeps `__proto__` as an ordinary own property rather than mutating the query object's prototype.

## Tests

- Added a regression test that removes `Object.fromEntries` and verifies a polling handshake still succeeds.
- `npx mocha --bail --exit test/server.js --grep 'without Object.fromEntries'`
- `npx prettier --check lib/server.ts lib/userver.ts test/server.js`
- `npx mocha --bail --exit` reached 90 passing tests (including the new regression and existing `__proto__` coverage); it then hit an existing xhr timeout after `npm ci --ignore-scripts` left the optional HTTP/3 native module unavailable on Windows.

### DIFF
--- a/packages/engine.io/lib/server.ts
+++ b/packages/engine.io/lib/server.ts
@@ -166,6 +166,23 @@ function parseSessionId(data: string): string | undefined {
   } catch (e) {}
 }
 
+export function parseQuery(params: URLSearchParams): Record<string, string> {
+  const query: Record<string, string> = {};
+
+  params.forEach((value, key) => {
+    // Define instead of assigning so a `__proto__` query parameter is kept as
+    // a regular own property, matching Object.fromEntries().
+    Object.defineProperty(query, key, {
+      value,
+      enumerable: true,
+      configurable: true,
+      writable: true,
+    });
+  });
+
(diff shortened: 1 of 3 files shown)


darrachequesne:

This should be fixed by https://github.com/socketio/socket.io/commit/ae7fb46e08c5ed964b4a1ea8b1703e816511598e.

Issue #3095, opened 2017-10-21
Socket still fires event after disconnect


wagenaartje:
### You want to:

* [x] report a *bug*

### Current behaviour
I'm sending certain tasks to connected clients (all using Chrome), the clients then send back the result of the task. Of course, all tasks need to be completed and thus a task should be sent to another client when the first client disconnects.

So this is what is happening:

- SERVER sends task to CLIENT
- CLIENT disconnects (fires `socket.on('disconnect', ...)`)
- SERVER removes CLIENT from the task's client list and appoints task to new CLIENT2
- CLIENT sends a message with task results (fires `socket.on('completion', ...)`)

I have no idea how this is even possible. I have spent literally days trying to understand why the server would freeze and I just discovered this by doing the following:

```js
  socket.on('completion', function (data) {
    console.log('Completing task...', socket.connected); // logs 'false'
    // do something
  });
```

### Steps to reproduce (if the current behaviour is a bug)

I can't really give 'steps'. All I can say is that i'm sending tasks to the client that are finished within a second, and i'm sending 1000s of them sequentially. I should also mention that each client has X sockets doing tasks, one for every CPU core.

### Expected behaviour
Should not fire an event when disconnected.

### Setup
- OS: Windows 10
- browser: Chrome
- socket.io version: latest

xPaw:
I believe I found out why this happens.

If I'm reading the code right, it emits normal events in `process.nextTick` in `ondecoded` and then in `dispatch`, but it appears that `onclose` would be emitted instantly, so it can be 2 ticks behind?

Could any of the developers confirm that this is what is happening?

darrachequesne:
For future readers: this was (finally) fixed by https://github.com/socketio/socket.io/commit/494c64e44f645cbd24c645f1186d203789e84af0 and included in [`socket.io@3.1.2`](https://github.com/socketio/socket.io/releases/tag/3.1.2).

PR #5531, opened 2026-08-01
Fix event dispatch before disconnection

### LINKED ISSUE #5213: Events before disconnection dropped
**Describe the bug**  
In a Socket.IO application, events emitted just before a client disconnects are not properly handled by specific `on(...)` listeners. While the `onAny(...)` listener detects these events, the specific `on(...)` listeners do not. This issue occurs with **any event** emitted just before disconnection. The reason `onAny(...)` works is that it runs before the dispatch, which relies on `process.nextTick`.

**To Reproduce**  

Socket.IO server version: `4.x.x` (replace with your version)  

*Server*

```js

const io = new Server(3000, {});

  console.log(`connect ${socket.id}`);

  // Example event listener (applies to any event)
  socket.on("exampleEvent", () => {
    console.log(`exampleEvent received before disconnect ${socket.id}`);
  });

  // Using onAny listener
  socket.onAny((event, ...args) => {
    console.log(`onAny received ${event} before disconnect ${socket.id}`);
  });

  socket.on("disconnect", () => {
    console.log(`disconnect ${socket.id}`);
  });
});
```

Socket.IO client version: `4.x.x` (replace with your version)

*Client*

```js

const socket = io("ws://localhost:3000/", {});

socket.on("connect", () => {
  console.log(`connect ${socket.id}`);
  
  // Emit any event just before disconnecting
  socket.emit("exampleEvent");
  socket.disconnect();
});

socket.on("disconnect", () => {
  console.log("disconnect");
});
```

**Expected behavior**  
The `on` listener for any event (e.g., `exampleEvent`) should detect and handle the event before disconnection. Currently, the `onAny` listener detects the event, but the specific `on(...)` listener does not. This occurs because `onAny` runs before the dispatch, which relies on `process.nextTick`, allowing it to detect the event before disconnection.

**Platform:**  

Not relevant

**Additional context**  
This issue affects any event emitted just before disconnection. The `onAny` listener works because it operates before the event dispatch process, which uses `process.nextTick`. As a result, events are detected by `onAny` but missed by the specific `on(...)` listeners, which rely on the dispatch mechanism that happens after disconnection begins.

mjibulu:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

When a complete non-binary event is received before disconnection, `onAny()` can observe it while the named listener is skipped because dispatch is deferred until after the socket's connection state changes.

### New behavior

Complete non-binary events whose middleware completes synchronously are dispatched immediately while the socket is connected. Asynchronous middleware, events received after disconnection, and binary events retain their existing deferred and guarded behavior.

Fixes #5213

The existing binary-packet regression test associated with #3095 remains unchanged and passes.

### DIFF
--- a/packages/socket.io/lib/socket.ts
+++ b/packages/socket.io/lib/socket.ts
@@ -36,6 +36,23 @@ import {
 
 const debug = debugModule("socket.io:socket");
 
+function hasBinary(data: unknown): boolean {
+  if (!data || typeof data !== "object") {
+    return false;
+  }
+  if (
+    Buffer.isBuffer(data) ||
+    data instanceof ArrayBuffer ||
+    ArrayBuffer.isView(data)
+  ) {
+    return true;
+  }
+  if (Array.isArray(data)) {
+    return data.some(hasBinary);
+  }
+  return Object.values(data).some(hasBinary);
+}
+
 const RECOVERABLE_DISCONNECT_REASONS: ReadonlySet<DisconnectReason> = new Set([
   "transport error",
   "transport close",
@@ -571,7 +588,10 @@ export class Socket<
         listener.apply(this, args);
       }
     }
-    this.dispatch(args);
+    this.dispatch(
+      args,
+      packet.type === PacketType.BINARY_EVENT || hasBinary(args),
+    );
   }
 
   /**
@@ -817,10 +837,12 @@ export class Socket<
    * @param {Array} event - event that will get emitted
    * @private
    */
-  private dispatch(event: Event): void {
+  private dispatch(event: Event, isBinary: boolean): void {
     debug("dispatching an event %j", event);
+    let isSynchronous = true;
+
     this.run(event, (err) => {
-      process.nextTick(() => {
+      const invoke = () => {
         if (err) {
           return this._onerror(err);
         }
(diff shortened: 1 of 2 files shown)


packages/socket.io/lib/socket.ts
```diff
@@ -571,7 +588,10 @@ export class Socket<
         listener.apply(this, args);
       }
     }
-    this.dispatch(args);
+    this.dispatch(
+      args,
+      packet.type === PacketType.BINARY_EVENT || hasBinary(args),
```
darrachequesne, line 593: Could you please explain the usage of the hasBinary() check here? Both plaintext and binary events should be complete in the `ondecoded()` handler, no?

packages/socket.io/lib/socket.ts
```diff
@@ -817,10 +837,12 @@ export class Socket<
...
    * @private
    */
-  private dispatch(event: Event): void {
+  private dispatch(event: Event, isBinary: boolean): void {
     debug("dispatching an event %j", event);
+    let isSynchronous = true;
+
     this.run(event, (err) => {
-      process.nextTick(() => {
+      const invoke = () => {
```
darrachequesne, line 845: Couldn't we just remove the `process.nextTick()` call here? Since there is already a `nextTick()` in the `ondecoded` handler.

https://github.com/socketio/socket.io/blob/45873ca93e6e25ac55dc1ba5c9b87853bf19ffba/packages/socket.io/lib/client.ts#L290-L292

darrachequesne (commented):
Hi! Thanks for your pull request. Please find my review notes below.

PR #5537, opened 2026-08-17
fix(sio): parse Engine.v3 namespace query without Object.fromEntries

JiuYue0820:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

`socket.io` still advertises `engines.node: >=10.2.0`. Engine.v3 namespace handshake parsing in `Client._parseNamespace` uses `Object.fromEntries()` on `URLSearchParams`, which does not exist on Node 10/11.

A protocol-3 connect packet with query auth therefore throws on the Node versions the package still claims to support.

### New behavior

Parse the query with a local `objectFromEntries` helper that does not depend on `Object.fromEntries`.

Related leftover from #5527. Separate from the engine.io HTTP/uWS query parsers.

### DIFF
--- a/packages/socket.io/lib/client.ts
+++ b/packages/socket.io/lib/client.ts
@@ -11,6 +11,16 @@ import type { Socket as RawSocket } from "engine.io";
 
 const debug = debugModule("socket.io:client");
 
+function objectFromEntries(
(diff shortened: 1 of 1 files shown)


darrachequesne:
Thanks :+1:

PR #5539, opened 2026-08-22
fix(adapter): do not let volatile packets break connection state recovery

chuanghiduoc:
## Problem

When `connectionStateRecovery` is enabled, a volatile event whose last argument is a string breaks the recovery of the client that received it.

Reproduction (server + client from current main):

```js
const io = new Server(httpServer, { connectionStateRecovery: {} });

  // ...
});

// after a normal broadcast has initialized the offset chain:
io.volatile.emit("status", "connected");
```

The client then disconnects and reconnects: `socket.recovered` is `false`, and every packet emitted during the disconnection is silently lost.

With the same scenario minus the `io.volatile.emit(...)` line, `socket.recovered` is `true`.

## Cause

The two sides of the offset mechanism disagree about volatile packets:

- server side (`SessionAwareAdapter.broadcast()`): a volatile EVENT packet gets **no** offset appended
- client side (`emitEvent()`): any trailing string argument of a received event is stored as the offset

So after `io.volatile.emit("status", "connected")`, the client stores the literal string "connected" as its offset. On reconnection, `restoreSession()` does `findIndex(packet.id === "connected")`, which fails, and the session is dropped.

This also affects the cluster adapters, which share the same condition in `addOffsetIfNecessary()`. I have deliberately not touched them here, because their packet storage/replay lives in the external adapter packages and needs to be changed in sync.

## Fix

In the in-memory adapter, volatile EVENT packets are now part of the offset chain like any other EVENT packet (they get an offset, and they are stored), but they are excluded from the replay in `restoreSession()`. Their semantics are unchanged - a volatile packet is still never delivered late.

## Testing

- new test in `packages/socket.io-adapter/test/index.ts`: session restore still succeeds with the volatile packet's offset, and the volatile packet is still excluded from `missedPackets`
- `npm test` in `packages/socket.io-adapter`: 49 passing (48 before)
- full suite in `packages/socket.io`: 216 passing

I also verified end-to-end with a script against the built packages: before the fix `recovered: false` / no replay; after the fix `recovered: true` / non-volatile packets replayed in order / volatile packets not replayed.

### DIFF
--- a/packages/socket.io-adapter/lib/in-memory-adapter.ts
+++ b/packages/socket.io-adapter/lib/in-memory-adapter.ts
@@ -466,7 +466,9 @@ export class SessionAwareAdapter extends Adapter {
     const missedPackets = [];
     for (let i = index + 1; i < this.packets.length; i++) {
       const packet = this.packets[i];
-      if (shouldIncludePacket(session.rooms, packet.opts)) {
+      // volatile packets are part of the offset chain but are never replayed
+      const isVolatile = packet.opts.flags?.volatile !== undefined;
+      if (!isVolatile && shouldIncludePacket(session.rooms, packet.opts)) {
         missedPackets.push(packet.data);
       }
     }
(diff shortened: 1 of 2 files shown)


packages/socket.io-adapter/lib/in-memory-adapter.ts
darrachequesne, line 496: How about not storing `packet.data` if the event is volatile?

PR #5540, opened 2026-08-22
fix(engine.io): run the polling write callback when the client aborts a compressed response

chuanghiduoc:
## Problem

Since 91dd763 ("perf(eio): stream compressed polling responses"), `doWrite()` invokes its callback only from the response's `'finish'` handler. If the client aborts the polling request while a large compressed payload is still streaming, `'finish'` never fires — and nothing errors on the zlib pipeline either, because writes into the destroyed socket just become no-ops.

Consequences of the lost callback (`write()` in polling.ts):

- `req.cleanup()` never runs for that request
- `this.emit("drain")` never fires, so the send callbacks queued for that batch (`sentCallbackFn`) are dropped

Before 91dd763, the whole payload was buffered and the callback ran unconditionally right after `res.end()`, so this could not happen.

Reproduction: new test `packages/engine.io/test/compression-abort.js` — a client opens a polling request with `Accept-Encoding: gzip`, the server keeps sending incompressible random data (threshold 0), and the client destroys the request as soon as the response headers arrive. On current main, the server-side send callback of that batch is never invoked; with this PR it always is (verified by stashing the fix and re-running).

## Fix

The callback now also runs on the response's `'close'` event, with a guard so it still runs exactly once (`'close'` also fires after normal completion):

```ts
this.res.once("finish", done);
this.res.once("close", done);
```

## Testing

- new test: `packages/engine.io/test/compression-abort.js` (fails without the fix, passes with it)
- existing suites unaffected

### DIFF
--- a/packages/engine.io/lib/transports/polling.ts
+++ b/packages/engine.io/lib/transports/polling.ts
@@ -322,6 +322,15 @@ export class Polling extends Transport {
     const stream = compressionMethods[encoding](this.httpCompression);
 
     let isErrored = false;
+    let isDone = false;
+
+    const done = () => {
+      if (isDone || isErrored) {
+        return;
+      }
+      isDone = true;
+      callback();
+    };
 
     stream.on("error", (err) => {
       isErrored = true;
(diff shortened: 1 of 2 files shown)


darrachequesne:
Good catch, thanks :+1:

darrachequesne:
Released in `engine.io@6.6.10`.

Release notes: https://github.com/socketio/socket.io/releases/tag/engine.io%406.6.10

PR #5541, opened 2026-08-23
fix: allow session recovery even if no event was received

### LINKED ISSUE #5538: Connection State Recovery fails if client reconnects before server emits any events ("Zero-Event" reconnect)
**Describe the bug**  
When `connectionStateRecovery` is enabled, a client that connects, joins rooms, and disconnects **before the server emits any events** cannot recover its session on reconnection. 

Even if the client reconnects within `maxDisconnectionDuration`, `socket.recovered` evaluates to `false`, a new `socket.id` is generated, and room memberships / `socket.data` are lost. This happens because `_lastOffset` on the client is `undefined` in the zero event case, causing the server's `typeof offset === "string"` check in `namespace.ts` to fail and bypass `restoreSession()`.

**To Reproduce**

Socket.IO server version: `4.8.1`

*Server*

```js

const io = new Server(3000, {
  connectionStateRecovery: {
    maxDisconnectionDuration: 2 * 60 * 1000,
  },
});

io.on("connection", async (socket) => {
  if (socket.recovered) {
    console.log(`RECOVERED: ${socket.id} (rooms: ${Array.from(socket.rooms).join(", ")})`);
    return;
  }

  console.log(`INITIAL CONNECT: ${socket.id}`);
  await socket.join("room:alerts");
  // Deliberately no events emitted to the client
});
```

Socket.IO client version: `4.8.1`

*Client*

```js

const socket = io("http://localhost:3000", {
  reconnection: true,
});

socket.on("connect", () => {
  console.log(`connect ${socket.id} (recovered: ${socket.recovered})`);

  // Simulate a brief network drop shortly after connecting
  setTimeout(() => {
    socket.io.engine.close();
  }, 200);
});
```

**Expected behavior**  
When the client reconnects after the transient drop, the server should recognize the `pid`, restore the previous session (`socket.id`, room memberships, `socket.data`), and set `socket.recovered = true`.

**Platform:**
 - Device: PC / Mac / Server
 - OS: Linux / macOS / Windows
 - Node.js version: `20.x` / `22.x`

**Additional context**  
* **Root Cause:** In `socket.io-client`, `_lastOffset` remains `undefined` until a packet of type `EVENT` with an offset is received. On reconnect, the handshake sends `{ pid: "<id>", offset: undefined }`. In `socket.io/lib/namespace.ts`, recovery is gated by:
  ```typescript
  if (
    this.server.opts.connectionStateRecovery &&
    typeof sessionId === "string" &&
    typeof offset === "string"
  ) {
    session = await this.adapter.restoreSession(sessionId, offset);
  }
  ```
  Because `offset` is `undefined`, `restoreSession()` is skipped entirely despite having a valid active `pid`.

* **Impact:** Quiet subscriber channels (e.g. alert rooms, idle game lobbies) lose session state on brief network blips unless application code manually emits a synthetic dummy event on first connect to "prime" `_lastOffset`.

* **Potential Solutions:**
  1. Include the current server stream offset in the initial `CONNECT` (packet type `0`) handshake payload alongside `{ sid, pid }` so `_lastOffset` is initialized at connection time.
  2. Allow `restoreSession(sessionId, offset)` to run when `offset` is `undefined`, restoring `id`, `rooms`, and `data` without replaying events.

marwan562:

* [x] a bug fix

### Current behavior

When `connectionStateRecovery` is enabled, a client that connects, joins rooms, and disconnects **before the server emits any events** cannot recover its session on reconnection (`socket.recovered` is `false`, a new `socket.id` is generated, room memberships and `socket.data` are lost).

This happens because `_lastOffset` is only set on the client when an EVENT packet is received. In the "zero-event" case, the reconnect CONNECT packet contains no offset at all (`{"pid":"..."}`), and the server-side check in `namespace.ts`:

```ts
if (
  this.server.opts.connectionStateRecovery &&
  typeof sessionId === "string" &&
  typeof offset === "string"
) {
```

skips `restoreSession()` entirely.

### New behavior

The offset is now optional in `Adapter#restoreSession()`:

- in `namespace.ts`, a non-string offset is normalized to `undefined`, and recovery is attempted as long as a valid private session id (`pid`) is provided
- in `SessionAwareAdapter#restoreSession()`, an undefined offset means the client did not receive any event yet, so the session (id, rooms, data) is restored without replaying any event

Note for external adapters (Redis adapter, Postgres adapter, cluster adapters...): `restoreSession()` may now be called with `offset === undefined`. Implementations that don't handle it can keep their current signature; returning `null` preserves the previous behavior.

Fixes #5538

Testing:
- added 2 cases in the server raw-packet suite (`test/connection-state-recovery.ts`): zero-event recovery (id/rooms/data restored, no replayed events) and recovery with a non-string offset
- added 1 case in `packages/socket.io-adapter/test/index.ts`: restore with an undefined offset returns the session with an empty `missedPackets` array
- added 1 end-to-end case in `packages/socket.io-client/test/connection-state-recovery.ts`: real client reconnects with `recovered === true` without any prior event
- all suites pass locally: `npm test -w packages/socket.io` (217 passing), adapter suite (22 passing), client node suite (112 passing; one pre-existing flaky failure in the unrelated `autoUnref` child-process test, which also fails on a clean checkout)

### DIFF
--- a/packages/socket.io-adapter/lib/in-memory-adapter.ts
+++ b/packages/socket.io-adapter/lib/in-memory-adapter.ts
@@ -387,11 +387,11 @@ export class Adapter extends EventEmitter {
   /**
    * Restore the session and find the packets that were missed by the client.
    * @param pid
-   * @param offset
+   * @param offset - the offset of the last packet received by the client, if any
    */
   public restoreSession(
     pid: PrivateSessionId,
-    offset: string,
+    offset?: string,
   ): Promise<Session> {
     return null;
   }
@@ -444,7 +444,7 @@ export class SessionAwareAdapter extends Adapter {
 
   override restoreSession(
     pid: PrivateSessionId,
-    offset: string,
+    offset?: string,
   ): Promise<Session> {
     const session = this.sessions.get(pid);
     if (!session) {
@@ -458,16 +458,18 @@ export class SessionAwareAdapter extends Adapter {
       this.sessions.delete(pid);
       return null;
     }
-    const index = this.packets.findIndex((packet) => packet.id === offset);
-    if (index === -1) {
-      // the offset may be too old
-      return null;
-    }
     const missedPackets = [];
-    for (let i = index + 1; i < this.packets.length; i++) {
-      const packet = this.packets[i];
-      if (shouldIncludePacket(session.rooms, packet.opts)) {
-        missedPackets.push(packet.data);
+    if (offset !== undefined) {
+      const index = this.packets.findIndex((packet) => packet.id === offset);
+      if (index === -1) {
+        // the offset may be too old
+        return null;
+      }
+      for (let i = index + 1; i < this.packets.length; i++) {
+        const packet = this.packets[i];
+        if (shouldIncludePacket(session.rooms, packet.opts)) {
+          missedPackets.push(packet.data);
+        }
       }
     }
     return Promise.resolve({
--- a/packages/socket.io-adapter/lib/in-memory-adapter.ts
+++ b/packages/socket.io-adapter/lib/in-memory-adapter.ts
@@ -458,18 +458,19 @@ export class SessionAwareAdapter extends Adapter {
       this.sessions.delete(pid);
       return null;
     }
-    const missedPackets = [];
+    let index = -1;
     if (offset !== undefined) {
-      const index = this.packets.findIndex((packet) => packet.id === offset);
+      index = this.packets.findIndex((packet) => packet.id === offset);
       if (index === -1) {
         // the offset may be too old
         return null;
       }
-      for (let i = index + 1; i < this.packets.length; i++) {
-        const packet = this.packets[i];
-        if (shouldIncludePacket(session.rooms, packet.opts)) {
-          missedPackets.push(packet.data);
-        }
+    }
+    const missedPackets = [];
+    for (let i = index + 1; i < this.packets.length; i++) {
+      const packet = this.packets[i];
+      if (shouldIncludePacket(session.rooms, packet.opts)) {
+        missedPackets.push(packet.data);
       }
     }
     return Promise.resolve({
--- a/packages/socket.io-adapter/test/index.ts
+++ b/packages/socket.io-adapter/test/index.ts
@@ -551,5 +551,48 @@ describe("socket.io-adapter", () => {
 
       expect(session).to.be(null);
     });
+
+    it("should restore a known session without offset", async () => {
+      const adapter = new SessionAwareAdapter({
+        server: {
+          encoder: {
+            encode(packet) {
+              return packet;
+            },
+          },
+          opts: {
+            connectionStateRecovery: {
+              maxDisconnectionDuration: 5000,
+            },
+          },
+        },
+      });
+
+      adapter.persistSession({
+        sid: "abc",
+        pid: "def",
+        data: "ghi",
+        rooms: ["r1", "r2"],
+      });
+
+      adapter.broadcast(
+        {
+          nsp: "/",
+          type: 2,
+          data: ["hello"],
+        },
+        {
+          rooms: new Set(),
+          except: new Set(),
+        },
+      );
+
+      const session = await adapter.restoreSession("def");
+
+      expect(session).to.not.be(null);
+      expect(session.sid).to.eql("abc");
+      expect(session.pid).to.eql("def");
+      expect(session.missedPackets).to.eql([]);
+    });
   });
 });
--- a/packages/socket.io-adapter/test/index.ts
+++ b/packages/socket.io-adapter/test/index.ts
@@ -592,6 +592,39 @@ describe("socket.io-adapter", () => {
       expect(session).to.not.be(null);
       expect(session.sid).to.eql("abc");
       expect(session.pid).to.eql("def");
+      // when offset is undefined (zero-event case), replay should start from the beginning
+      expect(session.missedPackets.length).to.eql(1);
+      expect(session.missedPackets[0][0]).to.eql("hello");
+    });
+
+    it("should restore a known session without offset and without missed packets", async () => {
+      const adapter = new SessionAwareAdapter({
+        server: {
+          encoder: {
+            encode(packet) {
+              return packet;
+            },
+          },
+          opts: {
+            connectionStateRecovery: {
+              maxDisconnectionDuration: 5000,
+            },
+          },
+        },
+      });
+
+      adapter.persistSession({
+        sid: "abc",
+        pid: "def",
+        data: "ghi",
+        rooms: ["r1", "r2"],
+      });
+
+      const session = await adapter.restoreSession("def");
+
+      expect(session).to.not.be(null);
+      expect(session.sid).to.eql("abc");
+      expect(session.pid).to.eql("def");
       expect(session.missedPackets).to.eql([]);
     });
   });
--- a/packages/socket.io-client/test/connection-state-recovery.ts
+++ b/packages/socket.io-client/test/connection-state-recovery.ts
@@ -27,4 +27,30 @@ describe("connection state recovery", () => {
       });
     });
   });
+
+  it("should restore session even if no event was received", () => {
+    return wrap((done) => {
+      const socket = io(BASE_URL, {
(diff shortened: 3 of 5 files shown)


official-burak (changes requested):
The missing-offset path restores the session identity but drops the packets that recovery is supposed to replay.

`SessionAwareAdapter#restoreSession` currently walks `this.packets` from the last offset the client processed. A client that never received an EVENT has no offset. That is not the same as "there were no packets." It means there is no last-processed id, so replay should start at the beginning of the buffer (still filtered by `shouldIncludePacket`).

With this PR, `offset === undefined` skips the loop entirely and always returns `missedPackets: []`. The sequence the feature exists for then loses events:

1. Client connects, joins a room, receives no EVENT, so it has no offset.
2. Client disconnects (`persistSession`).
3. Server emits to that room (stored in `this.packets`).
4. Client reconnects with `{ pid }` only.

Rooms and `socket.data` come back, `socket.recovered` is true, and the event from step 3 is gone. The new adapter test encodes that: it `broadcast`s `"hello"` and then asserts `missedPackets` is empty. That assertion is the bug.

Suggested shape:

```ts
let index = -1;
if (offset !== undefined) {
  index = this.packets.findIndex((packet) => packet.id === offset);
  if (index === -1) {
    return null; // unknown / too old offset, same as today
  }
}
for (let i = index + 1; i < this.packets.length; i++) {
  // existing shouldIncludePacket filter
}
```

Missing offset: `index` stays `-1`, loop starts at 0. Unknown string offset: still fail closed.

Please do not coerce a non-string `auth.offset` to `undefined`. `{"pid":"...","offset":123}` is malformed, not "no event yet." Treat only a missing offset as the zero-event case. A number or object should skip recovery (or fail the restore), not silently restore without replay. The new `"offset":123` test currently locks in that coercion.

A cleaner alternative, if you would rather keep `restoreSession(pid, offset: string)`: give the client an initial offset on CONNECT even when no EVENT has been sent, so `_lastOffset` is always a string. Either way, recovery without an offset still has to replay packets emitted after `persistSession`.

The namespace change to attempt restore when only `pid` is present looks right. The adapter replay is what needs to change, and the tests should cover "no offset + event during the disconnect window => packet is replayed."

marwan562:
Hi @official-burak — thanks for the thorough review, you're absolutely right.

**Understanding the issue:**
- The current `offset === undefined` path in `SessionAwareAdapter#restoreSession` incorrectly returns `missedPackets: []`, which drops events that were emitted while the client was disconnected. A missing offset means "no last-processed id", not "no packets to replay", so replay should start from index 0 (filtered by `shouldIncludePacket`).
- Coercing a non-string `auth.offset` (e.g. `{"pid":"...","offset":123}`) to `undefined` is wrong — malformed offset should fail closed (skip recovery / return null), not silently restore as a zero-event case.
- The new adapter test `broadcast + restore without offset => []` encodes the bug, and the `offset:123 => recovered true` test locks in the coercion bug. The namespace change to allow restore with only `pid` is correct, but the adapter replay needs fixing and we need a test for `no offset + event during disconnect => packet is replayed`.

**Planned fixes:**
1. **Adapter (`packages/socket.io-adapter/lib/in-memory-adapter.ts`)** — use suggested shape:
   ```ts
   let index = -1;
   if (offset !== undefined) {
     index = this.packets.findIndex(p => p.id === offset);
     if (index === -1) return null;
   }
   for (let i = index+1; i < this.packets.length; i++) { // starts at 0 when offset is undefined
     if (shouldIncludePacket(...)) missedPackets.push(...)
   }
   ```
2. **Namespace (`packages/socket.io/lib/namespace.ts`)** — don't coerce non-string offset. Only treat *missing* offset as zero-event:
   ```ts
   const offset = auth.offset;
   if (opts.connectionStateRecovery && typeof pid === "string" && (typeof offset === "string" || offset === undefined)) {
     // offset may be undefined, else malformed => skip recovery
   }
   ```
3. **Tests:**
   - Fix `socket.io-adapter/test/index.ts:555` to expect replayed packet when `offset` undefined (instead of `[]`).
   - Replace `offset:123 => recovered true` test in `packages/socket.io/test/connection-state-recovery.ts` with a failure case (`recovered false`, new id).
   - Keep zero-event recovery test for empty buffer, and add new test `no offset + emit during disconnect => packet replayed` (and adapter equivalent).

Will push an updated commit shortly. Let me know if the alternative approach (initial offset on CONNECT) would be preferred instead — happy to switch — but for now I'll fix the replay path as suggested.

marwan562:
Pushed fix addressing the review — thanks again for catching the replay bug.

**Changes in 27c686a:**

* **`SessionAwareAdapter#restoreSession`** (`packages/socket.io-adapter/lib/in-memory-adapter.ts:458`): now follows the suggested shape `let index = -1; if (offset !== undefined) { index = findIndex; if (index===-1) return null; } for (i=index+1 ...)`. When `offset` is `undefined` (zero-event case) replay starts at `0`, so `io.to(room).emit` while the client is disconnected is no longer lost. Unknown string offsets still fail closed.

* **`Namespace#_createSocket`** (`packages/socket.io/lib/namespace.ts:387`): no longer coerces `auth.offset` to `undefined`. Only `undefined` (missing) or a `string` is accepted: `typeof pid === "string" && (typeof offset === "string" || offset === undefined)`. Malformed offsets like `123` now skip recovery (new `sid`, `recovered:false`) instead of silently restoring.

* **Tests:**
  - `socket.io-adapter/test/index.ts:555` fixed to expect `missedPackets[0][0] === "hello"` when restoring without offset (was incorrectly `[]`), plus new test for empty buffer without missed packets.
  - `socket.io/test/connection-state-recovery.ts`: replaced `offset:123 => recovered:true` with `should not restore if offset not a string` (expects `recovered:false`, new id), kept the empty-buffer zero-event test, and added `should replay missed packets even if client did not receive any event before disconnection` (emits `hello` + `broadcast` during disconnect and asserts both are replayed before `CONNECT`).

All suites pass locally: `socket.io-adapter` 50 passing, `socket.io` 219 passing. Let me know if you'd prefer the alternative approach (initial offset on CONNECT) or any docs update for external adapters — happy to adjust.

official-burak (approved):
Replay from a missing offset looks right now, and a non-string offset fails closed.

marwan562:
Thanks for the review and the catch on the replay issue. Glad it looks right.

darrachequesne:

However, I'm worried that changing `restoreSession()` to accept `undefined` would alter the adapter contract and could break third-party implementations.

First-party adapters would need to be updated too:

- MongoDB adapter: https://github.com/socketio/socket.io-mongo-adapter/blob/eff10ab63ecfd5753d7985f46fd4278058f8570b/lib/index.ts#L804-L806

- Redis streams adapter: https://github.com/socketio/socket.io-redis-streams-adapter/blob/e4e38eb1065b04383749347f97b6442f60249a24/lib/adapter.ts#L443-L445

Since connection state recovery is explicitly best effort and currently based on pid + offset (see [1]), I think this change should be included in a future major release, if there is enough demand.

[1]: https://socket.io/docs/v4/connection-state-recovery#how-it-works-under-the-hood

marwan562:
Hi @darrachequesne, thanks for pointing this out and for reviewing.

You are right to worry about the adapter contract. I checked the two adapters you linked and I see the issue. Both the mongo adapter and redis streams adapter currently expect a string offset and will reject if it is undefined, so without an update they would just fail to recover the session. In my current PR the namespace catches that error and falls back to a new session, so it does not crash the server, but it also means zero event recovery would still not work with those adapters until they are updated.

I am happy to help update them if we go this way. For the mongo adapter that would mean skipping the ObjectId check when offset is undefined and fetching from the start, and for redis streams starting the XRANGE from the beginning of the stream. The in memory adapter already handles this. I can open companion PRs for those two adapters together with this one.

If you would rather keep the current signature as `offset: string` to avoid any contract change, I can also switch to the other approach @official-burak suggested and send an initial offset in the CONNECT packet so `_lastOffset` is always a string. That would avoid making the param optional.

Let me know which direction you prefer and I will push the update quickly. I would love to keep this fix for issue #5538 alive since it affects quiet rooms where the session is lost on a quick reconnect, and I am happy to do the extra work needed.

Thanks again for the guidance.

PR #5546, opened 2026-08-27
fix(client): discard timed-out buffered packets

### LINKED ISSUE #4318: Timed out event is still buffered
**Describe the bug**

For a timed out event, If connection between client and server is disrupted but not closed yet, client still buffers event.

**To Reproduce**

1- After connection between client and server is established disconnect the network.
2- before the client closes the connection send an event with timeout.
3- after the event times out connect the network.
4- client still sends the timed out event 

Socket.IO server version: `4.4.1`

*Server*

```js

const port = 3000;
const io = new Server(port);
console.log(`listening to ${port}`);

	console.log('user connected');
	socket.on("emited", (cb) => {
        console.log('got event');
		cb('done');
	});
});
});
```

Socket.IO client version: `4.4.1`

*Client*

```js

const socket = io("ws://server-ip:3000");

// send a message to the server
setTimeout(() => {
	socket.timeout(1000).emit("emited", (err, msg) => {
		if (err) {
			console.log(err);
			return;
		}
		console.log(msg);
	});
	console.log('event sent');
}, 10000);
```

**Expected behavior**

After the event times out, the event should have been removed from buffer so it wouldn't be sent to server.

**Platform:**
 - OS:Ubuntu 20.04

wangxpych:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

An event with an acknowledgement timeout can already be queued in the Engine.IO write buffer when the transport becomes temporarily unwritable. The Socket.IO acknowledgement callback times out and the packet is removed from the Socket.IO send buffer, but the Engine.IO packet remains queued and is sent when the transport becomes writable again.

For binary events, one Socket.IO event can map to several Engine.IO packets.

### New behavior

When an acknowledgement times out, the client removes every Engine.IO packet created from that Socket.IO event if none of them has been handed to the transport yet.

If any encoded packet is already part of the current transport batch, all related packets are preserved so a binary event cannot be left partially transmitted.

Fixes #4318.

Tests cover both multi-packet binary events and the partially-flushed safety boundary.

Verification:

- Engine.IO client formatting and TypeScript compilation
- Socket.IO client formatting and TypeScript compilation
- Engine.IO buffer-removal regression tests
- Socket.IO timeout tests (20 passing)

AI assistance disclosure: OpenAI Codex assisted with investigation, implementation, and test review; the change was manually validated.

### DIFF
--- a/packages/engine.io-client/lib/socket.ts
+++ b/packages/engine.io-client/lib/socket.ts
@@ -690,6 +690,26 @@ export class SocketWithoutUpgrade extends Emitter<
   }
 
   /**
+   * Removes packets that have not been handed to the transport yet.
+   *
+   * @param options - the options object shared by the packets to remove
+   * @private
+   */
+  public _removeFromWriteBuffer(options: WriteOptions) {
+    for (let i = 0; i < this._prevBufferLen; i++) {
+      if (this.writeBuffer[i].options === options) {
+        return;
+      }
+    }
+
+    for (let i = this.writeBuffer.length - 1; i >= this._prevBufferLen; i--) {
+      if (this.writeBuffer[i].options === options) {
+        this.writeBuffer.splice(i, 1);
+      }
+    }
+  }
+
+  /**
    * Flush write buffers.
    *
    * @private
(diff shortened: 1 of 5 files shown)


wangxpych:
Withdrawing this contribution after re-evaluating the repository's recent human review and merge activity. No action is needed from maintainers. Thank you.

PR #5547, opened 2026-08-28
Workaround to avoid false positives when detecting an offset

anas-861:
The last element of the arguments of a message received by the client can be a string even without an offset, for example in the case of sending only an event name or a message that contains a payload which is a string instead of an object, so the best approach is checking if the last argument is a string and follows the format of an offset. However, this approach is still prone to causing the same effect as before so I suggest designing a way of tracking if an offset was added or not. This workaround will prevent most errors where the frontend will try to use something that isn't an offset as if it were, causing an exception at the backend.

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior
If the client disconnects and the connection state recovery feature is enabled, in the case of the last message received not having an offset but having the last argument being a string will cause the front-end to assume it is an offset thus causing an exception at the back-end.

### New behavior
It will avoid most false positives by checking if the last string follows the "number-number" format (which is used for the offsets). However, this is still prone to false positives so tracking if an offset was added or not is suggested.

### DIFF
--- a/packages/socket.io-client/lib/socket.ts
+++ b/packages/socket.io-client/lib/socket.ts
@@ -772,7 +772,9 @@ export class Socket<
       }
     }
     super.emit.apply(this, args);
-    if (this._pid && args.length && typeof args[args.length - 1] === "string") {
+
+    // Consider designing a better way of tracking if an offset was actually added or not
+    if (this._pid && args.length && typeof args[args.length - 1] === "string" && /^\d+-\d+$/.test(args[args.length - 1]) ) {
       this._lastOffset = args[args.length - 1];
     }
   }


darrachequesne:
Hi! Thanks for the pull request. However, the format of the offset is not fixed, it is dependent on the adapter implementation:

- Redis streams: the stream entry ID (which indeed matches `^\d+-\d+$`)

https://github.com/socketio/socket.io-redis-streams-adapter/blob/e4e38eb1065b04383749347f97b6442f60249a24/lib/adapter.ts#L325-L330

- in-memory adapter: usage of `yeast()`

https://github.com/socketio/socket.io/blob/ae7fb46e08c5ed964b4a1ea8b1703e816511598e/packages/socket.io-adapter/lib/in-memory-adapter.ts#L486

- MongoDB: `ObjectId.toString("hex")`

https://github.com/socketio/socket.io-mongo-adapter/blob/eff10ab63ecfd5753d7985f46fd4278058f8570b/lib/index.ts#L508

Edit: we could maybe add a `args.length > 1` check though. What do you think?

PR #5550, opened 2026-08-29
fix(eio): report oversized polling payloads

### LINKED ISSUE #706: maxHttpBufferSize violation silently disconnects client with no error event
### Summary

When a client sends a message that exceeds the `maxHttpBufferSize`, the server closes the connection immediately with no way for the developer to detect or handle the error properly.

### Expected Behavior

There should be a proper `'error'` event (e.g., `'error'`, `'packetError'`, or `'close'` with a specific reason), or a packet with error metadata that allows distinguishing "oversized payload" from a general transport error.

### Actual Behavior

- Server does NOT trigger `socket.on('big-message')`
- `socket.conn.on('error')` is NOT called
- Only `socket.on('disconnect')` is called with reason `'transport error'`
- No detailed error message available

### Why this matters

This makes it **impossible to distinguish between an oversized payload vs. a broken connection**, which causes:
- Inaccurate logging / monitoring
- Inability to handle abuse cases
- Poor developer experience when debugging

### Minimal Repro

```ts
// Server (Node.js)
const io = new Server(server, {
  maxHttpBufferSize: 1e6
});
io.on('connection', (socket) => {
  socket.on('big-message', (data) => {
    console.log('Received:', data.length);
  });
  socket.on('disconnect', (reason) => {
    console.log('Disconnected:', reason);
  });
});

// Client (Browser or Node)
const socket = io('http://localhost:3000', { reconnection: false });
const big = 'x'.repeat(2 * 1024 * 1024); // 2MB
socket.emit('big-message', big);

wangxpych:
## Summary

- emit a transport error when an HTTP polling payload exceeds `maxHttpBufferSize`
- preserve the existing HTTP 413 response
- assert that server-side close handlers receive an actionable `TransportError`

## Context

The polling transport returned HTTP 413 when the incoming body crossed the configured limit, but it did not notify the Engine.IO socket about the cause. As a result, server-side Socket.IO handlers only received `transport close` without a description.

The uWebSockets polling implementation already reports `payload too large`, and the WebSocket implementation reports the corresponding `ws` range error. This change brings the Node.js polling transport in line with those paths by using the existing transport error pipeline after the request has been cleaned up.

Fixes socketio/engine.io#706

## Verification

- `npm run format:check --workspace=engine.io`
- Engine.IO Node test suite: 175 passing, 2 pending
- Engine.IO v3 compatibility suite: 175 passing, 2 pending
- regression test fails before the fix with `transport close` and passes after the fix with `transport error` / `payload too large`

The aggregate `npm test --workspace=engine.io` command reaches 175 passing tests but the local WebTransport test times out because the optional `@fails-components/webtransport-transport-http3-quiche` native module is not built in this environment.

AI assistance disclosure: OpenAI Codex assisted with reproduction, implementation, and verification. The change was reviewed and tested against the repository's current main branch.

### DIFF
--- a/packages/engine.io/lib/transports/polling.ts
+++ b/packages/engine.io/lib/transports/polling.ts
@@ -153,6 +153,7 @@ export class Polling extends Transport {
         exceededMaxHttpBufferSize = true;
         res.writeHead(413).end();
         cleanup();
+        this.onError("payload too large");
         return;
       }
 
--- a/packages/engine.io/test/server.js
+++ b/packages/engine.io/test/server.js
@@ -1823,7 +1823,7 @@ describe("server", () => {
       });
     });
 
-    it("should not be receiving data when getting a message longer than maxHttpBufferSize when polling", (done) => {
+    it("should report a message longer than maxHttpBufferSize when polling", (done) => {
       const opts = {
         allowUpgrades: false,
         transports: ["polling"],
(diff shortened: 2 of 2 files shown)


wangxpych:
Withdrawing this contribution after re-evaluating the repository's recent human review and merge activity. No action is needed from maintainers. Thank you.

PR #5557, opened 2026-09-12
test(packet): add unit test suite for packet payload buffer size validation

gcoinstash-cmd:
### Summary
Adds isolated unit tests for packet payload buffer size clamping:
- Valid payload byte bounds [0, maxPayload]
- Upper bound rejection
- Negative / NaN input rejection


darrachequesne:
Hi! Thanks for the pull request. Could you please explain the usage of the `validatePacketPayloadSize()` method?

PR #5560, opened 2026-09-20
fix(client): discard timed-out packets from the Engine.IO buffer

### LINKED ISSUE #4318: Timed out event is still buffered

dyk1454683243-sudo:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

An event emitted with `timeout()` while the connection is still considered open can already be sitting in the Engine.IO write buffer (transport temporarily unwritable, for example after a network disruption that has not closed the socket yet).

When the acknowledgement times out, Socket.IO removes the packet from its own `sendBuffer`, but the encoded Engine.IO packet remains queued. Restoring the network / making the transport writable again flushes that timed-out event to the server.

This is #4318.

### New behavior

On acknowledgement timeout the client also removes every Engine.IO packet created from that Socket.IO event, as long as none of them has been handed to the transport yet.

If any encoded packet is already part of the current flush batch, all related packets are left intact so a binary event cannot be partially transmitted.

Neighboring in-flight events are still flushed normally.

Fixes #4318.

Regression coverage:

- Engine.IO write-buffer removal, including the partially-flushed safety boundary
- Socket.IO timeout: timed-out event is not delivered after the transport becomes writable again (the original one/two/three scenario)
- Socket.IO timeout: binary events that encode to multiple Engine.IO packets


packages/engine.io-client/lib/socket.ts
```diff
@@ -689,6 +689,34 @@ export class SocketWithoutUpgrade extends Emitter<
...
+   *
+   * @param options - the options object shared by the packets to remove
+   * @private
+   */
+  /* private */ _removeFromWriteBuffer(options: WriteOptions) {
+    if (options == null) {
+      return;
+    }
+
+    for (let i = 0; i < this._prevBufferLen; i++) {
```
darrachequesne, line 707: I think there is a edge case here, in case of a Socket.IO binary packet sent as multiple Engine.IO packets:

- the first packet is sent successfully
- "drain" is emitted and `_prevBufferLen` is set to 0
- the attachments remains in the Engine.IO buffer and are then removed by `_removeFromWriteBuffer()`
- the server-side Socket.IO parser is left waiting for attachments that will never arrive

darrachequesne (commented):
Hi! Thanks a lot for the pull request, please find my comments below.

darrachequesne:
Alternative fix to https://github.com/socketio/socket.io/issues/4318:

- Socket.IO assigns discard metadata to the Engine.IO packets generated for a timeout-based emit.
- All Engine.IO packets belonging to the same logical Socket.IO packet share a `discardGroupId`.
- The group has a `discardAfter` timestamp.
- Before Engine.IO flushes its write buffer, it drops expired groups atomically, as long as no fragment from that group has already been handed to the transport.

Something like:

```typescript
interface WriteOptions {
  compress?: boolean;
  discardGroupId?: number;
  discardAfter?: number;
}
```

Which would keep Engine.IO as the owner of its own write buffer and avoid Socket.IO calling into a new Engine.IO buffer API.

What do you think?

Related: https://github.com/socketio/socket.io/pull/5546

PR #5561, opened 2026-09-22
fix(sio): isolate recovery offsets across child namespaces

xingsy97:

* [x] a bug fix
* [ ] a new feature
* [ ] other

### Current behavior

With connection state recovery enabled, room-targeted parent broadcasts pass the same packet to every child adapter. Each child appends its recovery offset to the shared data array, also modifying events already cached by earlier children. A replayed event can therefore carry another child's offset, causing a subsequent session recovery to fail.

### New behavior

Shallow-copy the packet and its data array for each child adapter. Recovery offsets remain isolated, while nested payload objects and Buffers retain their original references.


darrachequesne:

I wonder if it wouldn't be better to fix this in the `addOffsetIfNecessary` method here: https://github.com/socketio/socket.io/blob/3df3fed075fd2312e84481786235aa14d08b8fe7/packages/socket.io-adapter/lib/cluster-adapter.ts#L471-L473

This way, a new array will be created only if connection state recovery is enabled.

What do you think?

